Issue #629, opened 2023-04-25
issues when installing from PyPI


hasan7n:
1. Following the documentation (https://mlcommons.github.io/GaNDLF/setup/#install-from-package-managers), trying to install using `pip install gandlf` gives the following error:
```
ERROR: Could not find a version that satisfies the requirement gandlf (from versions: 0.0.1, 0.0.16.dev0, 0.0.16.dev20230321, 0.0.16.dev20230322, 0.0.16.dev20230323, 0.0.16.dev20230324, 0.0.16.dev20230325, 0.0.16.dev20230326, 0.0.16.dev20230327, 0.0.16.dev20230328, 0.0.16.dev20230329, 0.0.16.dev20230330, 0.0.16.dev20230331, 0.0.16.dev20230401, 0.0.16.dev20230402, 0.0.16.dev20230403, 0.0.16.dev20230404, 0.0.16.dev20230405, 0.0.16.dev20230406, 0.0.16.dev20230407, 0.0.16.dev20230408, 0.0.16.dev20230409, 0.0.16.dev20230410, 0.0.16.dev20230411, 0.0.16.dev20230412, 0.0.16.dev20230413, 0.0.16.dev20230414, 0.0.16.dev20230415, 0.0.16.dev20230416, 0.0.16.dev20230417, 0.0.16.dev20230418, 0.0.16.dev20230419, 0.0.16.dev20230420, 0.0.16.dev20230421, 0.0.16.dev20230422, 0.0.16.dev20230423, 0.0.16.dev20230424, 0.0.16.dev20230425)
ERROR: No matching distribution found for gandlf
```
running `pip install gandlf==0.0.16.dev20230425`, for example, works.

2. Installing using `conda` doesn't seem to work. Took more than 10 minutes waiting for this:
```
Collecting package metadata (current_repodata.json): done
Solving environment: / 
```
Then I stopped it.

3. When installed using PyPI, `gandlf_deploy` doesn't work properly. I think it is because it tries to search for the Dockerfiles. The build context is being misconfigured (set to `/home/hasan_proj5/anaconda3/envs/gantest/lib/python3.8/site-packages` in my case)

AlexanderGetka-cbica:
Regarding point 3: I am looking into this.

This issue arises from the fact that the structure of the repo and the structure of the installed package are different.
The Dockerfiles we use for CI/CD assume that the build context will be at the top level of the repo and that setup.py and all other files needed for installing the package will be available.
[...]

PR #643, opened 2023-05-10
Fix deployment to work when installed from wheel

AlexanderGetka-cbica:
Attempts to fix one aspect of #629.

WARNING: The changes needed to make this work without restructuring the entire repo are pretty wacky, bear with me when reading the below.

Deployment was failing from GaNDLF installed via wheel/from pip largely because the directory structure in this case is NOT equivalent to the source repo's structure (i.e., from `pip install -e .` in the git repo). 
Trying to install the dockerfiles/other files needed for setup into the installed package is difficult for various reasons related to implementation details in setuptools, etc.
In order to avoid restructuring the entire repository, I decided to leave these files in the top-level site-packages directory, and "install them" during deployment. 
This means turning the GaNDLF package directory into something that, itself, can be installed.
So we use some tricks that require us to exclude the other stuff inside the package from being installed, including making a recursive symlink to the GaNDLF package so that the code gets bundled, but change setup.py to exclude the package from being installed by itself. (Isn't recursion fun?)

## Proposed Changes

- deploy script now checks for the case where GaNDLF is installed directly as a package in site-packages
- We now bundle the extra setup-related files and install to site-packages
- deploy script now uses these to generate the install inside the container

## Checklist

- [x] I have read the [`CONTRIBUTING`](https://github.com/mlcommons/GaNDLF/blob/master/CONTRIBUTING.md) guide.
- [x] My PR is based from the [current GaNDLF master ](https://garygregory.wordpress.com/2016/11/10/how-to-catch-up-my-git-fork-to-master/).
- [x] Function/class source code documentation added/updated.
- [x] Code has been [blacked](https://github.com/psf/black#usage) for style consistency.
- [x] [History](https://github.com/mlcommons/GaNDLF/blob/master/HISTORY.md) has been updated, if appropriate.


hasan7n:
@AlexanderGetka-cbica how are you testing this?

I `pip install`ed gandlf from source and tested that deployment works fine. But do you know how I can "simulate" having it installed just like `pip install gandlf` from pypi?

I did `pip install . -t <some_folder>` and visually inspected that folder. I can see that it is installed according to your description above. So this looks good, but any idea if there is a better way for testing?

AlexanderGetka-cbica:
In my case, I am just generating a wheel, and installing from that. the Python "build" tool can do this, e.g. from the source repo:

pip install build
python -m build

(you might be able to get it to build just the wheel and not also the source distribution). This is also how we create the nightly builds. 

Then install the wheel directly
pip install GaNDLF_[...]_py38.whl

AlexanderGetka-cbica:
I am not sure how testing could cover this. We could make the testing pipeline cover the entire build->upload->install->run tests, but i don't know if pytest even supports this use.

hasan7n (dismissed):
Thanks @AlexanderGetka-cbica ! It works

PR #841, opened 2024-04-02
Updated version checking mechanism

### LINKED ISSUE #840: [BUG] Moving from version `0.0.XX` to `0.Y.0` creates compatibility issues.
### Describe the bug
When the new CLI is ready (#818), we will transition to version `0.1.0`. However, this is creating issues with the [`version_check`](https://github.com/mlcommons/GaNDLF/blob/2b2445cba24e98b3dd51dd6645324ed5d5cd2769/GANDLF/utils/generic.py#L106) function.

### To Reproduce
1. Put min/max version in the config as `0.0.XX` and `0.1.0`.
2. Run any training command.
3. See version compatibility error.

### Expected behavior
Clearly, `0.1.0` is greater than `0.0.XX`. However, the custom [`version_check`](https://github.com/mlcommons/GaNDLF/blob/2b2445cba24e98b3dd51dd6645324ed5d5cd2769/GANDLF/utils/generic.py#L106)  cannot seem to comprehend this.

### Media
N.A.

### Environment information

```
GANDLF version: 0.0.20-dev
Git hash: 3c61892f9d5ffc9d17720ccfd413c1b23cfd50f0
Platform: Windows-10-10.0.22631-SP0
Machine: AMD64
Processor: Intel64 Family 6 Model 165 Stepping 2, GenuineIntel
Architecture: 64bit WindowsPE
Python environment:
  Version: 3.11.6
  Implementation: CPython
  Compiler: MSC v.1935 64 bit (AMD64)
  Build: main Oct  3 2023 10:29:11
```

### Additional context
N.A.

scap3yvt:
Fixes #840

## Proposed Changes

- `version_check` updated to use the [`Packaging` library](https://packaging.pypa.io/en/latest/development/index.html).
- added checks for `minimum` and `maximum` version
- updated `sys.exit` with `assert`

## Checklist

- [x] [`CONTRIBUTING`](https://github.com/mlcommons/GaNDLF/blob/master/CONTRIBUTING.md) guide has been followed.
- [x] PR is based on the [current GaNDLF master ](https://docs.github.com/en/desktop/contributing-and-collaborating-using-github-desktop/keeping-your-local-repository-in-sync-with-github/syncing-your-branch-in-github-desktop?platform=windows).
- [x] Function/class source code documentation added/updated (ensure `typing` is used to provide type hints, including and not limited to using `Optional` if a variable has a pre-defined value).

### DIFF
--- a/GANDLF/utils/generic.py
+++ b/GANDLF/utils/generic.py
@@ -1,12 +1,14 @@
+from typing import Optional, Union
 import os, datetime, subprocess, sys
 from copy import deepcopy
 import random
+from packaging.version import Version
+
 import numpy as np
 import torch
 import SimpleITK as sitk
 from contextlib import contextmanager, redirect_stderr, redirect_stdout
 from os import devnull
-from typing import Optional, Union
 
 
 @contextmanager
@@ -86,23 +88,6 @@ def get_filename_extension_sanitized(filename: str) -> str:
     return ext
 
 
-def parse_version(version_string: str) -> int:
-    """
-    Parses version string, discards last identifier (NR/alpha/beta) and returns an integer for comparison.
-
-    Args:
-        version_string (str): The string to be parsed.
-
-    Returns:
-        int: The version number.
-    """
-    version_string_split = version_string.replace("-dev", "")
-    version_string_split = version_string_split.split(".")
-    if len(version_string_split) > 3:
-        del version_string_split[-1]
-    return int("".join(version_string_split))
-
-
 def version_check(version_from_config: str, version_to_check: str) -> bool:
     """
     This function checks if the version of the config file is compatible with the version of the code.
@@ -114,15 +99,21 @@ def version_check(version_from_config: str, version_to_check: str) -> bool:
     Returns:
         bool: If the version of the config file is compatible with the version of the code.
     """
-    version_to_check_int = parse_version(version_to_check)
-    min_ver = parse_version(version_from_config["minimum"])
-    max_ver = parse_version(version_from_config["maximum"])
-    if (min_ver > version_to_check_int) or (max_ver < version_to_check_int):
-        sys.exit(
-            "Incompatible version of GaNDLF detected ("
-            + str(version_to_check_int)
-            + ")"
-        )
+    version_to_check_int = Version(version_to_check)
+    min_ver = Version(version_from_config.get("minimum"), None)
+    max_ver = Version(version_from_config.get("maximum"), None)
+
+    assert min_ver is not None, "Minimum version is not specified in the config file"
+    assert max_ver is not None, "Maximum version is not specified in the config file"
+    assert (
+        min_ver <= max_ver
+    ), "Minimum version is greater than maximum version in the config file"
+    assert (
+        min_ver <= version_to_check_int
+    ), "Minimum version requested in config is greater than the GaNDLF version"
+    assert (
+        max_ver >= version_to_check_int
+    ), "Maximum version requested in config is less than the GaNDLF version"
 
     return True
 
--- a/setup.py
+++ b/setup.py
@@ -113,6 +113,7 @@ requirements = [
     "zarr",
     "keyring",
     "monai==1.3.0",
+    "packaging==24.0",
 ]
 
 if __name__ == "__main__":


GANDLF/utils/generic.py
```diff
@@ -86,23 +88,6 @@ def get_filename_extension_sanitized(filename: str) -> str:
...
-        int: The version number.
-    """
-    version_string_split = version_string.replace("-dev", "")
-    version_string_split = version_string_split.split(".")
-    if len(version_string_split) > 3:
-        del version_string_split[-1]
-    return int("".join(version_string_split))
-
-
 def version_check(version_from_config: str, version_to_check: str) -> bool:
```
VukW, line 91: Can you plz fix typing here also (is it dict, right?)
```suggestion
def version_check(version_from_config: dict[str, str], version_to_check: str) -> bool:
```

GANDLF/utils/generic.py
```diff
@@ -114,15 +99,21 @@ def version_check(version_from_config: str, version_to_check: str) -> bool:
...
-    max_ver = parse_version(version_from_config["maximum"])
-    if (min_ver > version_to_check_int) or (max_ver < version_to_check_int):
-        sys.exit(
-            "Incompatible version of GaNDLF detected ("
-            + str(version_to_check_int)
-            + ")"
-        )
+    version_to_check_int = Version(version_to_check)
+    min_ver = Version(version_from_config.get("minimum"), None)
+    max_ver = Version(version_from_config.get("maximum"), None)
```
VukW, line 104: What is `None` as second arg here? don't see any second arg in the doc

VukW (changes requested):

GANDLF/utils/generic.py
```diff
@@ -114,15 +99,21 @@ def version_check(version_from_config: str, version_to_check: str) -> bool:
...
-    version_to_check_int = parse_version(version_to_check)
-    min_ver = parse_version(version_from_config["minimum"])
-    max_ver = parse_version(version_from_config["maximum"])
-    if (min_ver > version_to_check_int) or (max_ver < version_to_check_int):
-        sys.exit(
-            "Incompatible version of GaNDLF detected ("
-            + str(version_to_check_int)
-            + ")"
-        )
+    version_to_check_int = Version(version_to_check)
```
VukW, line 102: And also rename variable plz as it is not `int` anymore

GANDLF/utils/generic.py
```diff
@@ -114,15 +99,21 @@ def version_check(version_from_config: str, version_to_check: str) -> bool:
...
+    min_ver = Version(version_from_config.get("minimum"), None)
+    max_ver = Version(version_from_config.get("maximum"), None)
+
+    assert min_ver is not None, "Minimum version is not specified in the config file"
```
VukW, line 106: `min_ver` and `max_ver` cannot be None: they either a valid `Version` object, or execution already failed on a previous line if `version_from_config.get("minimum")` is not a valid version string

GANDLF/utils/generic.py
```diff
@@ -114,15 +99,21 @@ def version_check(version_from_config: str, version_to_check: str) -> bool:
...
+    assert max_ver is not None, "Maximum version is not specified in the config file"
+    assert (
+        min_ver <= max_ver
+    ), "Minimum version is greater than maximum version in the config file"
+    assert (
+        min_ver <= version_to_check_int
+    ), "Minimum version requested in config is greater than the GaNDLF version"
+    assert (
+        max_ver >= version_to_check_int
+    ), "Maximum version requested in config is less than the GaNDLF version"
```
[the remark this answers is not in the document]
scap3yvt: Thanks, this makes sense. Done.

GANDLF/utils/generic.py
VukW: Hm
I see you imply `version_from_config["minimum"]` can be int. Is it really so? Looks like neither old code (`parse_version` function) nor new (`Version`) cannot parse ints:
```
In [4]: Version(1)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[4], line 1
----> 1 Version(1)
....
    197 if not match:
    198     raise InvalidVersion(f"Invalid version: '{version}'")

TypeError: expected string or bytes-like object
```

GANDLF/utils/generic.py
VukW: :+1:
Then you may simplify this: you don't need to pass `None` as default value explicitly:
```
In [5]: d = {"a": 2, "b": 123}

In [6]: d.get("c")

In [7]: str(d.get("c"))
Out[7]: 'None'

In [8]: d.get("c") is None
Out[8]: True
```

scap3yvt: This has changed. The `None` should be the default for `.get()`. Thanks for catching it!

GANDLF/utils/generic.py
VukW: 🤔 I was talking about assert check actually:) it would never be raised as `min_ver` variable is always not a None

scap3yvt: Ah yes, I understand now. Done.

GANDLF/utils/generic.py
VukW: @scap3yvt this `Version(version_from_config.get("minimum", None))`, `.get` + `None` are excessive: if `minimum` version was not found in the config, you pass `None` and the code anyway fails because `Version()` cannot process it. So no need to guard from empty config as version is obligatory any way.
So, you may simplify it to just
```
min_ver = Version(version_from_config["minimum"])
max_ver = Version(version_from_config["maximum"])
```

scap3yvt: Got it, and I have made appropriate changes.

VukW: Thanks a lot, man!

VukW (approved):
Thank you for the fixes! Getting rid of custom code in favor of standard solutions really makes it look better:)

PR #845, opened 2024-04-09
New command line API functionality

sarthakpati:
Fixes #ISSUE_NUMBER

## Major Changes

- Usability updates
  - new CLI functionality to make it standardized using `click`, and related changes to workflows and tests
  - standardized logging (#893)
  - torch version update (#857)
  - workflows updated to use specific `pip` versions (#887)
  - single function to get all augmentations (#889)
  - allow metrics functionality to use separate target and prediction inputs (#900)
  - trigger CI tests for all branches (#855)
  - added check for case-insensitive headers for metrics (https://github.com/mlcommons/GaNDLF/pull/845/commits/22ee9bdf47ac22cbb01af18d4e68dde402b26f4f)
- New features
  - new segmentation architecture (#873)
  - new metrics (#885)

## Checklist

- [x] [`CONTRIBUTING`](https://github.com/mlcommons/GaNDLF/blob/master/CONTRIBUTING.md) guide has been followed.
- [x] PR is based on the [current GaNDLF master ](https://docs.github.com/en/desktop/contributing-and-collaborating-using-github-desktop/keeping-your-local-repository-in-sync-with-github/syncing-your-branch-in-github-desktop?platform=windows).
- [x] Function/class source code documentation added/updated (ensure `typing` is used to provide type hints, including and not limited to using `Optional` if a variable has a pre-defined value).


VukW:
Hi folks!
We have a couple steps before PR can be merged:
1. fix & merge https://github.com/mlcommons/GaNDLF/pull/853 where some output param names are changed. @scap3yvt 
2. Despite all the new code is covered by tests, the overall coverage was dropped as we included entrypoint scripts to be calculated for coverage. We can either add some new tests for that or decrease coverage threshold. @sarthakpati  what do you think?

sarthakpati:
Apologies, somehow I missed this comment, @VukW!


> 1. fix & merge [Updated CLI options for new API #853](https://github.com/mlcommons/GaNDLF/pull/853) where some output param names are changed. @scap3yvt

This should be done.

> 2. Despite all the new code is covered by tests, the overall coverage was dropped as we included entrypoint scripts to be calculated for coverage. We can either add some new tests for that or decrease coverage threshold. @sarthakpati  what do you think?

How difficult would it be to add new tests? To note that this branch is supposed to stay as a separate branch (and not get merged into master) for at least a few months.

VukW:
> To note that this branch is supposed to stay as a separate branch (and not get merged into master) for at least a few months.

@sarthakpati 😱Why so? It's not a big problem to add a new test, but it may be disturbing to support two branches (this and main) simultaneously - especially if we'd need to fix / improve anything in entrypoints scripts (that's highly probable on the range of a few months)

sarthakpati:
I agree (and the idea to delay the merge is more of a strategic endeavor than anything else). Do you think it makes sense to move the master the newer API and keep a "legacy" branch? I am honestly agnostic to either, just that we need to delay the actual release of the new API branch for a few months.

VukW:
Can you plz describe the strategic reason a bit wider as I don't get it quite well? I mean, depending on why exactly we want to wait without merging, I am imaginating different possible solutions. But as this PR is not breaking old behavior, in most of cases merging it looks better for me rather then supporting two branches. Different options I do see:

a. Merge as-is. We do not break `old-way` behavior however we distantiate from supporting it, noting that in future all new improvements and features would be implemented in new-way CLI.
b. Mark new-way CLI as `experimental` and old-day scripts as `stable` and merge. Thus, users would still use old-versioned entrypoint scripts, but it would be much easier for us to implement new features both in old-way and new-way commands.
c. Postpone merging for a couple of weeks..one month if there are any specific issues / notes that we want to introduce together with a newer CLI. 
d. Decline the whole PR / new CLI API


Say, we have the following confusions:
1. _"We are introducing breaking changes to API? Wow, great! Then lets also fix X and Y and Z and maybe smth else right now, we don't want our users to change their scripts one more time in the future!"._
  Say, standartizing input/output param names. IMO best solutions are `(c)` (postpone for a short time if we already know what should be fixed) or `(b)` (add new API as experimental and play with it to understand what can be improved)
2. _"The whole PR looks like too complex... What if we break something? What if we break user behavior?"_
  As PR is designed as non-breaking, we ensure users can still run old scripts as previously (in particular, unit tests assure it), so we may do `(a)`. However if we still unconfident, we can do `(b)`, marking new-way as experimental.
3. _"New CLI code looks more complex for developers, it would be harder to understand it for new developers"_
  Not arguing with subjectivity / objectivity of the reason (anyway, it's just an example), there might be cases when we are confused with the basic foundation of this PR. In such a case, `(d)` - declining everything - looks like a reasonable solution for stopping over-engineering.
4. _"If merging new CLI, then we need to bump version to 1.0, but this requires to refactor all other code"_.
  In this case it seems to me more feasible to eat it piece by piece - say, merge new CLI `(a)` as `0.1.0`, but not bumping major version (`1.0.0`) until everything is fixed 

So, for me it seems like merging the new CLI or declining it totally is always better (from the terms of supportability) then keeping two branches. And if keep two branches - it might be possible, but for some finite period of time until we decide what to do further.
That's why I'm asking to understand better a specific reason why / what do we want to do with this in future:)

sarthakpati:
@VukW: Wow, that was elaborate 😄 

Anyway, the reason behind having a separate branch was to delay the actual "release" (i.e., into a tag) of the new API for a few months. But your assessment is completely spot-on. Considering we are a tight knit group of contributors, I vote for option `(a)`, but after #842 has been merged in. The reason behind that is that it would be great to have a holistic logging functionality available along with the new APIs. Thoughts, @sylwiamm?

VukW:
@sarthakpati 👍 So, we are not pushing new release / new tag / new wheel package for now, but we kindly ask for all new PRs to be based on this branch instead of master, right? In this way it's easier for us to merge any new features / fixes rather then merge them to master & solve conflicts after

sarthakpati:
> So, we are not pushing new release / new tag / new wheel package for now, but we kindly ask for all new PRs to be based on this branch instead of master, right?

Yes, precisely! But I would suggest let's wait for @sylwiamm to wrap up the logging work and then we proceed with this. Thoughts?

sarthakpati:
Hey @VukW, now that #893 has been merged, shall we merge this in and then point users who need the older API to https://github.com/mlcommons/GaNDLF/tree/older_apis?

VukW:
Hmmm
So, I do see the following tags right now:
- `0.0.19` tag, made in Feb 2024
- current `master`, version `0.0.20-dev` with a bunch of fixes over `0.0.19`
- `old_branches` == `master` + a deprecation notice in Readme
- this branch, aimed to be merged to master and make a `0.1.0` version

Current master is in stable and working state, right? What do you think if we make two separate releases sequentially: first, release current master under `0.0.20` version, and then merge and release this new-api branch under `0.1.0` version?
In this case people who depend on `v0.0.19` would be able to update up to `0.0.20` automatically (without any backward incompatibilities) and to use some added features.
Thus, we won't need to support a separate branch (that doesn't really match gandlf's default git flow) - instead we just mark current `master` with a `0.0.20` tag and that's aligned with all the previous and future releases

*If you are not sure master is stable enough right now, we can use your proposed solution - just keep this `old_apis` branch, without actual releasing it

sarthakpati:
That's a great idea, thank you! I am going to get started on this.

sarthakpati:
@mlcommons/gandlf-write - there are multiple folks who have contributed to this PR (including all maintainers). Any proposals on what protocol to follow WRT reviews before merging?

GANDLF/config_manager.py
```diff
@@ -618,9 +620,7 @@ def _parseConfig(
             params["model"]["class_list"] = temp_classList.split(",")
         else:
             try:
-                params["model"]["class_list"] = ast.literal_eval(
-                    params["model"]["class_list"]
-                )
+                params["model"]["class_list"] = eval(params["model"]["class_list"])
             except AssertionError:
```
szmazurek, line 624: Is it correct error type to rise here? AssertionError could happen only in response to failed evaluation of assert statement I think. If so, then it will never be risen

GANDLF/config_manager.py
```diff
@@ -738,4 +738,10 @@ def ConfigManager(
     Returns:
         dict: The parameter dictionary.
     """
-    return _parseConfig(config_file_path, version_check_flag)
+    try:
+        return _parseConfig(config_file_path, version_check_flag)
+    except Exception as e:
+        logging.info(
```
szmazurek, line 744: Minor thing, but perhaps we should use logging.error here to indicate errors? The log is easier to read

GANDLF/entrypoints/collect_stats.py
```diff
@@ -129,7 +135,78 @@ def plot_all(df_training, df_validation, df_testing, output_plot_dir):
...
+    df_testing["split"] = "test"
+    df_validation["split"] = "validation"
+    pd.concat((df_training, df_testing, df_validation)).to_csv(output_file)
+
+
+def _collect_stats(model_dir: str, output_dir: str):
+    input_dir = os.path.normpath(model_dir)
+    output_dir = os.path.normpath(output_dir)
+    Path(output_dir).mkdir(parents=True, exist_ok=True)
+    # TODO: here is bug? output_file is never used. We don't save data.csv at all?
```
szmazurek, line 164: is this TODO relevant? I see `output_file` is being read

GANDLF/config_manager.py
sarthakpati: Agreed - changed to `logging.error`

GANDLF/config_manager.py
sarthakpati: Done.

GANDLF/entrypoints/collect_stats.py
sarthakpati: Agreed.

GANDLF/entrypoints/run.py
```diff
@@ -0,0 +1,291 @@
...
+            train_mode=train_flag,
+            device=device,
+            resume=resume_flag,
+            reset=reset_flag,
+            output_dir=output_path,
+        )
+    except Exception:
+        # TODO: why so? Why not just default way when exception is printed
+        #  and process exits with code 1 automatically?
+        sys.exit("ERROR: " + traceback.format_exc())
```
szmazurek, line 80: As now GaNDLF uses logging, maybe here it would be also worth to log the Error and traceback before exit

GANDLF/entrypoints/run.py
```diff
@@ -0,0 +1,291 @@
...
+    "-d",
+    # TODO: Not sure it's worth to restrict this list. What about other devices?
+    #  GaNDLF guarantees to work properly with these two options, but
+    #  other values may be partially working also.
+    #  * GaNDLF code convert `-1` to `cpu` (i.e. it is expected somebody may pass -1)
+    #  * `cuda:0` should work also, isn't it? Just would not be treated as `cuda`
+    #  * Would `mps` work?
+    #  * int values (like `1`) - are they supported? (legacy mode for cuda https://pytorch.org/docs/stable/tensor_attributes.html#torch-device)
+    type=click.Choice(["cuda", "cpu"]),
+    required=True,  # FIXME: either keep default value, or set required flag
```
szmazurek, line 126: Maybe defaulting here to one of the values depending on availability is worth considering?

GANDLF/entrypoints/run.py
```diff
@@ -0,0 +1,291 @@
...
+        "-d",
+        "--device",
+        default="cuda",  # TODO: default value doesn't work as arg is required
+        metavar="",
+        type=str,
+        required=True,
+        help="Device to perform requested session on 'cpu' or 'cuda'; for cuda, ensure CUDA_VISIBLE_DEVICES env var is set",
+    )
+    parser.add_argument(
+        "-rt",
```
szmazurek, line 238: maybe aslo reset and resume flags could use `store_true` flag to avoid confusing string evaluation? Although this is being deprecated, perhaps some old users may have some scripts done and are aware of that - open for discussion if this should be modiifed

GANDLF/entrypoints/recover_config.py
```diff
@@ -0,0 +1,104 @@
...
+    )
+
+    parser.add_argument(
+        "-m", "--modeldir", metavar="", type=str, help="Path to the model directory."
+    )
+    # TODO: despite of `str` type, real value is never used (only checks if it is filled or not)
+    #  Thus, caveats:
+    #    * passing `--mlcube False` would still process it as mlcube;
+    #    * passing `--mlcube "" ` (with empty str) acts as non-mlcube
+    parser.add_argument(
```
szmazurek, line 82: Similairly to previous ones - here `store_true` resolves the TODO problem

GANDLF/entrypoints/split_csv.py
```diff
@@ -0,0 +1,118 @@
...
+        default=None,
+        type=str,
+        required=True,
+        help="Output directory to save the split data.",
+    )
+
+    args = parser.parse_args()
+
+    # check for required parameters - this is needed here to keep the cli clean
+    for param_name in ["inputCSV", "outputDir", "config"]:
```
szmazurek, line 106: Is the thing in comment indeed true? Shouldn't flagging args as required=True be enough to ensure that they are passed?

GANDLF/entrypoints/verify_install.py
```diff
@@ -0,0 +1,56 @@
...
+        import GANDLF as gf
+
+        print("GaNDLF installed version:", gf.__version__)
+    except Exception as e:
+        raise Exception(
+            "GaNDLF not properly installed, please see https://mlcommons.github.io/GaNDLF/setup"
+        ) from e
+
+    # we always want to do submodule update to ensure any hash updates are ingested correctly
+    try:
```
szmazurek, line 23: Two questions - first I do not see anything git related here anymore, so perhaps comment is outdated. Secondly, conceptual question - should we indeed try to install the Gandlf for user if the version check fails? Shouldn't we just throw an error and ask to manually do it? Automated install here can go in some undesired ways (although as of right now cannot think of any). Thing to discuss, perhaps I am paranoid

GANDLF/metrics/classification.py
```diff
@@ -25,6 +23,36 @@ def overall_stats(
...
+        """
+        Convert the input tensor to integer format.
+
+        Args:
+            input_tensor (torch.Tensor): The input tensor.
+
+        Returns:
+            torch.Tensor: The tensor converted to integer format.
+        """
+        return_tensor = input_tensor
```
szmazurek, line 36: It needs to be checked but I believe such assignment still modifies input tensor inplace. I suppose if we want to avoid this side effect here deepcopy would be needed

GANDLF/metrics/classification.py
```diff
@@ -25,6 +23,36 @@ def overall_stats(
...
+        return return_tensor
+
+    # this is needed for a few metrics
+    # ensure that predictions and target are in integer format
+    prediction_wrap = __convert_tensor_to_int(prediction)
+    target_wrap = __convert_tensor_to_int(target)
+
+    # this is needed for auroc
+    # ensure that predictions are in integer format
+    prediction_wrap = prediction
```
szmazurek, line 48: Same as above, I believe the predictions themselves are going to be modified

GANDLF/metrics/classification.py
```diff
@@ -25,6 +23,36 @@ def overall_stats(
...
+            return_tensor = return_tensor.long()
+        return return_tensor
+
+    # this is needed for a few metrics
+    # ensure that predictions and target are in integer format
+    prediction_wrap = __convert_tensor_to_int(prediction)
+    target_wrap = __convert_tensor_to_int(target)
+
+    # this is needed for auroc
+    # ensure that predictions are in integer format
```
szmazurek, line 47: Is it checked somehow? I believe this is not necessairly true regarding the preds being int format. The idea of auroc itself is to check the model true and false positive rates in different decision thresholds, therefore any thresholding is possible only on logits, otherwise it does not make sense. User guide of torchmetrics also show example of raw logits https://lightning.ai/docs/torchmetrics/stable/classification/auroc.html

samples/config_getting_started_segmentation_rad3d.yaml
```diff
@@ -89,6 +89,6 @@ scheduler:
 track_memory_usage: false
 verbose: true
 version:
-  maximum: 0.0.20
-  minimum: 0.0.20
+  maximum: 0.1.0-dev
```
szmazurek, line 92: general comment regarding all configs with the min/max version specs - I do not know how it is done now, but if manually then perhaps some simple script doing grep and sed operations could do.

GANDLF/entrypoints/run.py
sarthakpati: Hmm, I am not sure if putting a default here makes a lot of sense. If someone is running inference on optimized models, they might prefer `cpu`, but if they are running training, they might prefer `cuda`. No?

GANDLF/entrypoints/run.py
sarthakpati: I am partial to not making major changes to the argeparse interfaces (since we will be deprecating them at a future release anyway)... Thoughts?

GANDLF/entrypoints/recover_config.py
sarthakpati: [Some comment](https://github.com/mlcommons/GaNDLF/pull/845#discussion_r1693723543).

GANDLF/entrypoints/split_csv.py
sarthakpati: Yeah, this is just to give a better error message. Doesn't make a lot of sense with `click`, but will no longer be needed once we deprecate this interface.

GANDLF/entrypoints/verify_install.py
sarthakpati: Agreed.

GANDLF/metrics/classification.py
sarthakpati: I needed to have this in place to get auroc to work for the BraTS-Path metrics (#885). Let's open up a separate issue to check this, instead?

samples/config_getting_started_segmentation_rad3d.yaml
sarthakpati: We have it: https://github.com/mlcommons/GaNDLF/blob/new-apis_v0.1.0-dev/update_version.py

GANDLF/metrics/classification.py
sarthakpati: Fixed.

GANDLF/metrics/classification.py
sarthakpati: Fixed.

GANDLF/config_manager.py
```diff
@@ -497,7 +499,7 @@ def _parseConfig(
                                 "max"
                             ] = sys.float_info.max
                 elif key in thresholdOrClipDict:
-                    sys.exit("Use only 'threshold' or 'clip', not both")
+                    logging.error("Use only 'threshold' or 'clip', not both")
```
szmazurek, line 502: Thanks for the update! Altough, should we completly disregard throwing an error here? My intention was to rather just log the error via logger in addition to throwing an error, be it assertion or sys.exit. Should we resign from this functionality? If so, then it is fine

GANDLF/config_manager.py
```diff
@@ -618,11 +620,11 @@ def _parseConfig(
...
         else:
             try:
-                params["model"]["class_list"] = ast.literal_eval(
-                    params["model"]["class_list"]
+                params["model"]["class_list"] = eval(params["model"]["class_list"])
+            except Exception as e:
+                logging.error(
+                    f"Could not evaluate the `class_list` in `model`, Exception: {str(e)}, {traceback.format_exc()}"
                 )
-            except AssertionError:
```
szmazurek, line 624: Same thing as above - should we disregard throwing error at all?

GANDLF/entrypoints/anonymizer.py
```diff
@@ -0,0 +1,134 @@
...
+from GANDLF.entrypoints import append_copyright_to_help
+from GANDLF.utils.gandlf_logging import logger_setup
+
+
+def _anonymize_images(
+    input_dir: str, output_file: str, config_path: Optional[str], modality: str
+):
+    input_dir = os.path.normpath(input_dir)
+    output_file = os.path.normpath(output_file)
+    # TODO: raise an error if config pass provided but not exist (user made a typo?)
```
szmazurek, line 23: Is todo still valid? If so, perhaps this check is worth adding now.

GANDLF/entrypoints/construct_csv.py
```diff
@@ -0,0 +1,181 @@
...
+    # This is used for MLCube functionality because MLCube does not support plain string inputs.
+    if channels_id.endswith(".yml") or channels_id.endswith(".yaml"):
+        if os.path.isfile(channels_id):
+            with open(channels_id, "r") as f:
+                content = yaml.safe_load(f)
+                channels_id = content["channels"]
+                if isinstance(channels_id, list):
+                    channels_id = ",".join(channels_id)
+
+                # TODO: raise a warning if label param is both passed as arg and defined in file
```
szmazurek, line 40: Is the todo still valid? If so, should the warning be added?

GANDLF/entrypoints/construct_csv.py
```diff
@@ -0,0 +1,181 @@
...
+        help="Label/mask identifier string to check for in all files in 'input_dir'; for example: --labelID _seg.nii.gz",
+    )
+    parser.add_argument(
+        "-o", "--outputFile", metavar="", type=str, help="Output CSV file"
+    )
+    parser.add_argument(
+        "-r",
+        "--relativizePaths",
+        metavar="",
+        type=ast.literal_eval,
```
szmazurek, line 158: should we also go there with `store_true` action instead of literal eval type? This would remove the need to write `True` or `False` as arg, but it is breaking change for previous api. If you think we should maintain compatibility this comment is irrelevant

GANDLF/entrypoints/run.py
```diff
@@ -0,0 +1,291 @@
...
+            train_mode=train_flag,
+            device=device,
+            resume=resume_flag,
+            reset=reset_flag,
+            output_dir=output_path,
+        )
+    except Exception:
+        # TODO: why so? Why not just default way when exception is printed
+        #  and process exits with code 1 automatically?
+        logging.error(f"ERROR: {traceback.format_exc()}")
```
szmazurek, line 80: Same regard as above, I believe we should still throw an exception - here also after logging program continues

GANDLF/entrypoints/run.py
szmazurek: Hmm, that's true, actually leaving it up to choice is not that big of a deal, at least user has to think about the device specification

GANDLF/entrypoints/run.py
szmazurek: Yeah, it would be a breaking change. If so, comment above is also irrelevant. Closing

GANDLF/entrypoints/split_csv.py
szmazurek: Agree!

GANDLF/metrics/classification.py
szmazurek: Yeah, good idea.

GANDLF/models/brain_age.py
```diff
@@ -27,7 +28,7 @@ def brainage(parameters):
         # Load the pretrained VGG16 model
         model = torchvision.models.vgg16(pretrained=True)
     except Exception:
-        sys.exit("Error: Failed to load VGG16 model: " + traceback.format_exc())
+        logging.error(f"Error: Failed to load VGG16 model: {traceback.format_exc()}")
```
szmazurek, line 31: Here as with the above replacements - I believe Exception should be still thrown

GANDLF/models/densenet.py
```diff
@@ -186,7 +186,7 @@ def __init__(self, parameters: dict, block_config=(6, 12, 24, 16)):
             self.output_size = (1, 1, 1)
             self.conv_stride = (parameters["conv1_t_stride"], 2, 2)
         else:
-            sys.exit("Only 2D or 3D convolutions are supported.")
+            logging.error("Only 2D or 3D convolutions are supported.")
```
szmazurek, line 189: Same here, exception should still be thrown even whe we are logging it

GANDLF/models/efficientnet.py
```diff
@@ -395,7 +395,7 @@ def __init__(self, parameters: dict, scale_params):  # how to scale depth and wi
         elif self.n_dimensions == 3:
             self.output_size = (1, 1, 1)
         else:
-            sys.exit("Only 2D or 3D convolutions are supported.")
+            logging.error("Only 2D or 3D convolutions are supported.")
```
szmazurek, line 398: Same as above, we should still throw error

szmazurek: Also when I think of it now, this check appears across many models (as it should). Perhaps it could be checked somewhere globally before params arrive to the model constructors.

GANDLF/models/resnet.py
```diff
@@ -52,7 +52,7 @@ def __init__(self, parameters: dict, blockType, block_config):
         elif self.n_dimensions == 3:
             self.output_size = (1, 1, 1)
         else:
-            sys.exit("Only 2D or 3D convolutions are supported.")
+            logging.error("Only 2D or 3D convolutions are supported.")
```
szmazurek, line 55: Same as in the comment of `efficientnet.py` file - we should throw an error.

GANDLF/config_manager.py
VukW: Hm. I was going to propose just raising an exception:
```suggestion
                    raise ValueError("Use only 'threshold' or 'clip', not both")
```
However I just noticed we don't log exceptions to the debug file (@benmalef FYI). I believe, we should fix that and then raising exception would be the cleanest solution - both logged & exiting the gandlf command with exit_code 1

sarthakpati: Brilliant, let's make it for the next PR before tagging the new API changes.

GANDLF/entrypoints/construct_csv.py
sarthakpati: We need to maintain compatibility with current API.

GANDLF/entrypoints/run.py
sarthakpati: Let's put this as a separate PR once @benmalef puts in exception-handling to the logging mechanism.

GANDLF/models/brain_age.py
sarthakpati: Same response - let's wait for the next PR.

sarthakpati:
@szmazurek I should have addressed all comments...

mlcube/metrics_mlcube/example_custom_entrypoint/getting_started_3d_rad_seg.py
```diff
@@ -40,29 +41,29 @@ def create_csv(predictions, labels):
...
+        output_file (str): The path to the output file/folder
+        config (str): The path to the parameters file
     """
     exit_status = os.system(
-        f"python3.9 gandlf_generateMetrics -c {parameters_file} -i ./data.csv -o {output_path}"
+        f"gandlf generate-metrics -c {config} -i ./data.csv -o {output_file}"
     )
     exit_code = os.WEXITSTATUS(exit_status)
-    sys.exit(exit_code)
+    logging.error(exit_code)
```
szmazurek, line 56: minor thing - are we sure it should be always logged as error? From what I understand this operation can be successfull too, so logging perhaps could depend on exit code value (i.e. info if 0 and error if 1)

mlcube/metrics_mlcube/example_custom_entrypoint/template.py
```diff
@@ -15,29 +16,29 @@ def create_csv(predictions, labels):
...
+        output_file (str): The path to the output file/folder
+        config (str): The path to the parameters file
     """
     exit_status = os.system(
-        f"python3.9 gandlf_generateMetrics -c {parameters_file} -i ./data.csv -o {output_path}"
+        f"gandlf generate-metrics -c {config} -i ./data.csv -o {output_file}"
     )
     exit_code = os.WEXITSTATUS(exit_status)
-    sys.exit(exit_code)
+    logging.error(exit_code)
```
szmazurek, line 31: Same as above - is this always valid to log with error level?

mlcube/model_mlcube/example_custom_entrypoint/getting_started_3d_rad_seg.py
```diff
@@ -22,32 +23,34 @@ def create_csv(data_path):
...
     exit_status = os.system(
-        "python3.9 gandlf_run --train False "
+        "gandlf run --infer "
         f"--device {device} --config /embedded_config.yml "
-        f"--modeldir /embedded_model/ -i ./data.csv -o {output_path}"
+        f"--model-dir /embedded_model/ -i ./data.csv -o {output_path}"
     )
     exit_code = os.WEXITSTATUS(exit_status)
-    sys.exit(exit_code)
+    logging.error(exit_code)
```
szmazurek, line 41: Same as above - is this always valid to log with error level?

mlcube/model_mlcube/example_custom_entrypoint/template.py
```diff
@@ -25,23 +26,23 @@ def run_gandlf(output_path, device):
...
     exit_status = os.system(
-        "python3.9 gandlf_run --train False "
+        "gandlf run --infer "
         f"--device {device} --config /embedded_config.yml "
-        f"--modeldir /embedded_model/ -i ./data.csv -o {output_path}"
+        f"--model-dir /embedded_model/ -i ./data.csv -o {output_path}"
     )
     exit_code = os.WEXITSTATUS(exit_status)
-    sys.exit(exit_code)
+    logging.error(exit_code)
```
szmazurek, line 34: Same as above - is this always valid to log with error level?

szmazurek (changes requested):
Super minor thing regarding logging levels and imo good to go

mlcube/metrics_mlcube/example_custom_entrypoint/getting_started_3d_rad_seg.py
```diff
@@ -40,29 +41,29 @@ def create_csv(predictions, labels):
...
+        output_file (str): The path to the output file/folder
+        config (str): The path to the parameters file
     """
     exit_status = os.system(
-        f"python3.9 gandlf_generateMetrics -c {parameters_file} -i ./data.csv -o {output_path}"
+        f"gandlf generate-metrics -c {config} -i ./data.csv -o {output_file}"
     )
     exit_code = os.WEXITSTATUS(exit_status)
-    sys.exit(exit_code)
+    logging.info(exit_code)
```
szmazurek, line 56:
```suggestion
    logging.info(exit_code) if exit_code == 0 else logging.error(exit_code)
```

szmazurek: I toguht of something like this - do you think it's fine or am I overkilling?

mlcube/metrics_mlcube/example_custom_entrypoint/template.py
```diff
@@ -15,29 +16,29 @@ def create_csv(predictions, labels):
...
+        output_file (str): The path to the output file/folder
+        config (str): The path to the parameters file
     """
     exit_status = os.system(
-        f"python3.9 gandlf_generateMetrics -c {parameters_file} -i ./data.csv -o {output_path}"
+        f"gandlf generate-metrics -c {config} -i ./data.csv -o {output_file}"
     )
     exit_code = os.WEXITSTATUS(exit_status)
-    sys.exit(exit_code)
+    logging.info(exit_code)
```
szmazurek, line 31:
```suggestion
    logging.info(exit_code) if exit_code == 0 else logging.error(exit_code)
```

mlcube/model_mlcube/example_custom_entrypoint/getting_started_3d_rad_seg.py
```diff
@@ -22,32 +23,34 @@ def create_csv(data_path):
...
     exit_status = os.system(
-        "python3.9 gandlf_run --train False "
+        "gandlf run --infer "
         f"--device {device} --config /embedded_config.yml "
-        f"--modeldir /embedded_model/ -i ./data.csv -o {output_path}"
+        f"--model-dir /embedded_model/ -i ./data.csv -o {output_path}"
     )
     exit_code = os.WEXITSTATUS(exit_status)
-    sys.exit(exit_code)
+    logging.info(exit_code)
```
szmazurek, line 41:
```suggestion
    logging.info(exit_code) if exit_code == 0 else logging.error(exit_code)
```

mlcube/model_mlcube/example_custom_entrypoint/template.py
```diff
@@ -25,23 +26,23 @@ def run_gandlf(output_path, device):
...
     exit_status = os.system(
-        "python3.9 gandlf_run --train False "
+        "gandlf run --infer "
         f"--device {device} --config /embedded_config.yml "
-        f"--modeldir /embedded_model/ -i ./data.csv -o {output_path}"
+        f"--model-dir /embedded_model/ -i ./data.csv -o {output_path}"
     )
     exit_code = os.WEXITSTATUS(exit_status)
-    sys.exit(exit_code)
+    logging.info(exit_code)
```
szmazurek, line 34:
```suggestion
    logging.info(exit_code) if exit_code == 0 else logging.error(exit_code)
```

szmazurek (changes requested):
Another remark regarding logging

mlcube/metrics_mlcube/example_custom_entrypoint/getting_started_3d_rad_seg.py
sarthakpati: I think/hope the mlcube mechanism will be changing soon, so this will most likely be redundant. 😄

mlcube/metrics_mlcube/example_custom_entrypoint/template.py
sarthakpati: [Same comment](https://github.com/mlcommons/GaNDLF/pull/845#discussion_r1697029762).

mlcube/model_mlcube/example_custom_entrypoint/getting_started_3d_rad_seg.py
sarthakpati: [Same comment](https://github.com/mlcommons/GaNDLF/pull/845#discussion_r1697029762).

mlcube/model_mlcube/example_custom_entrypoint/template.py
sarthakpati: [Same comment](https://github.com/mlcommons/GaNDLF/pull/845#discussion_r1697029762).

szmazurek (approved):

PR #881, opened 2024-06-04
Add logging - DEPRECATED

### LINKED ISSUE #755: Prettify stdout
The STDOUT can be full of warnings, it should be more user friendly

Make GaNLDF's stdout more user friendly (how?)

N/A

**Additional context**
Warnings could be useful debugging. Perhaps we can have an organized logging system: current stdout can be split into stdout for users, and debugging logs in a `~/.gandlf_logs` file for example.

benmalef:
Fixes #755
## Brief description
It adds the logging implementation. 

I have tried to implement these [requirements](https://github.com/mlcommons/GaNDLF/pull/842#discussion_r1570850957)

Please review the code and give me any feedback.. :)
It is a draft, it hasn't been completed yet.

---
log is splitted into four parts: debug log file, stdout, stderr, tqdm progress file

- **debug log:** file catches everything DEBUG+. In particular, it should catch errors and warnings etc also (thus, both stdout and stderr content is also catched by debug log file)
- **stdout:** displays INFO messages. Not lower, not higher
- **stderr:** displays WARNING+ messages
- **tqdm:** is not implemented. Maybe we can use [this](https://tqdm.github.io/docs/contrib.logging/)

---
### Usage
You can use the logger directly in the module:

Here is an example
``` 
from GANDLF.utils import gandlf_logger_setup
logger = gandlf_logger_setup(__name__) 
logger.warning("warning message") 
```

## Proposed Changes

- Create the gandlf_logger file in utils dir.
- Create the logging_config file.
- It contains gandlf_logger_setup fun that sets up the logger using the logging_config.yaml file.

## Checklist


GANDLF/compute/forward_pass.py
```diff
@@ -116,7 +118,7 @@ def validate_network(
         tqdm(valid_dataloader, desc="Looping over " + mode + " data")
     ):
         if params["verbose"]:
-            print("== Current subject:", subject["subject_id"], flush=True)
+            logger.debug("== Current subject:", subject["subject_id"])
```
VukW, line 121: Does it work? AFAIR logger does not support handy string concat as print does and requires you (developer) to format log message manually beforehand like 
```suggestion
            logger.debug('== Current subject: {subject["subject_id"]}')
```

GANDLF/compute/forward_pass.py
```diff
@@ -458,23 +460,20 @@ def validate_network(
...
+                for metric in params["metrics"]:
+                    if isinstance(total_epoch_valid_metric[metric], np.ndarray):
+                        to_print = (
+                            total_epoch_valid_metric[metric] / (batch_idx + 1)
+                        ).tolist()
+                    else:
+                        to_print = total_epoch_valid_metric[metric] / (batch_idx + 1)
+                    logger.debug(
+                        "Half-Epoch Average " + mode + " " + metric + " : ", to_print
+                    )
```
VukW, line 476:
```suggestion
                  for metric in params["metrics"]:
                      if isinstance(total_epoch_valid_metric[metric], np.ndarray):
                          to_print = (
                              total_epoch_valid_metric[metric] / (batch_idx + 1)
                          ).tolist()
                      else:
                          to_print = total_epoch_valid_metric[metric] / (batch_idx + 1)
                      logger.debug(
                          "Half-Epoch Average " + mode + " " + metric + " : ", to_print
                      )
```
You occasionally shifted this piece of code out from `if half-epoch` condition

logging_config.yaml
```diff
@@ -0,0 +1,37 @@
...
+handlers:
+  stdoutHandler: # only display info level
+    class: logging.StreamHandler
+    level: INFO
+    formatter: detailed
+    filters: [info_only_filter]
+    stream: ext://sys.stdout
+  stderrHandler: # display warning and above messages
+    class: logging.StreamHandler
+    level: ERROR
```
VukW, line 22:
```suggestion
    level: WARNING
```

Now stderr displays only errors, while warnings are hidden

logging_config.yaml
```diff
@@ -0,0 +1,37 @@
...
+    formatter: detailed
+    stream: ext://sys.stderr
+  debugHandler:
+    class: logging.FileHandler
+    filename: gandlf.log
+    formatter: detailed
+    level: DEBUG
```
VukW, line 29: Now `gandlf.log` is always appended, mixing logs from different runs. Is it ok? IMO after 10+ commands it would be hard to find any useful info there. Should we rewrite the file on every command? @benmalef @sarthakpati

VukW:
Hi @benmalef ! Thanks for the code. The first 4 points look exactly as expected, I personally like the elegancy of the configuration & usage :)

Concerning tqdm, your proposal with `logging_redirect_tqdm`  looks beauty but requires setting a context for every tqdm cycle. Let's think if we can simplify this by any kind of wrapper over `tqdm` ?.. so context is set up automatically when cycle is used...

VukW:
concerning tqdm: we can use the following wrapper to handle the redirect context automatically:
```
def custom_tqdm(it: Iterable, *args, **kwargs):
    with logging_redirect_tqdm():
        for x in tqdm(it, *args, **kwargs):
            yield x


for ic in custom_tqdm(range(10)):
    ....
```

benmalef:
Hi @VukW, 
First of all, thanks for your review and your comments. They are very meaningful. I will try to address them as soon as possible.

> Concerning tqdm, your proposal with `logging_redirect_tqdm` looks beauty but requires setting a context for every tqdm cycle. Let's think if we can simplify this by any kind of wrapper over `tqdm` ?.. so context is set up automatically when cycle is used...
>
I agree with you. 
I will try to do it. 
Maybe it would be better to create a new PR for this.?

benmalef:
> concerning tqdm: we can use the following wrapper to handle the redirect context automatically:
> 
> ```python
> def custom_tqdm(it: Iterable, *args, **kwargs):
>     with logging_redirect_tqdm():
>         for x in tqdm(it, *args, **kwargs):
>             yield x
> 
> 
> for ic in custom_tqdm(range(10)):
>     ....
> ```

Yes, we can do it like this. !! 
Thanks for the advice!!

logging_config.yaml
benmalef: Yes, it makes sense to me. 
Can we use the [rotating file handler](https://docs.python.org/3/library/logging.handlers.html#rotatingfilehandler) or simply rewrite the file ? @VukW

VukW: As we discussed with @sarthakpati here https://github.com/mlcommons/GaNDLF/pull/842#discussion_r1570850957 , lets make it rotating file handler somewhere in `/tmp/gandlf/` ?

benmalef:
Hi @VukW, @sarthakpati 
I am trying to solve this error in Openfl-test.  It can not find the logging_config.yaml file. 
Locally,  it is running fine without any errors.  Any thoughts ???
image

sarthakpati:
There is an error when running the mlcube test [[ref](https://productionresultssa18.blob.core.windows.net/actions-results/fb3ab814-af17-4a89-a0e5-3e1cecc62a10/workflow-job-run-c2d224cd-5211-5684-0e07-208199a0479e/logs/job/job-logs.txt?rsct=text%2Fplain&se=2024-06-18T17%3A36%3A03Z&sig=IORYowt68UISs86%2F5kL9nx%2BI0L%2FtTp7Ucx%2B3AgfS4EA%3D&ske=2024-06-19T03%3A08%3A19Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2024-06-18T15%3A08%3A19Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2023-11-03&sp=r&spr=https&sr=b&st=2024-06-18T17%3A25%3A58Z&sv=2023-11-03)] (which likely goes for others as well):

```python-traceback
2024-06-15T09:00:25.5466986Z ERROR: Traceback (most recent call last):
2024-06-15T09:00:25.5468631Z The ``converters`` are currently experimental. It may not support operations including (but not limited to) Functions in ``torch.nn.functional`` that involved data dimension
2024-06-15T09:00:25.5471978Z WARNING: Defining 'patch_sampler' as a string will be deprecated in a future release, please use a dictionary instead
2024-06-15T09:00:25.5491341Z   File "/home/runner/work/GaNDLF/GaNDLF/GANDLF/entrypoints/run.py", line 66, in _run
2024-06-15T09:00:25.5492557Z WARNING: Initializing 'memory_save_mode' as False
2024-06-15T09:00:25.5493179Z     main_run(
2024-06-15T09:00:25.5494396Z   File "/home/runner/work/GaNDLF/GaNDLF/GANDLF/cli/main_run.py", line 63, in main_run
2024-06-15T09:00:25.5495445Z WARNING: Initializing 'determinism' as False
2024-06-15T09:00:25.5496174Z WARNING: Initializing 'previous_parameters' as None
2024-06-15T09:00:25.5498525Z     logger = gandlf_logger_setup(__name__)
2024-06-15T09:00:25.5501608Z   File "/home/runner/work/GaNDLF/GaNDLF/GANDLF/utils/gandlf_logger.py", line 63, in gandlf_logger_setup
2024-06-15T09:00:25.5502722Z     with open(config_path, "r") as file:
2024-06-15T09:00:25.5503661Z FileNotFoundError: [Errno 2] No such file or directory: 'logging_config.yaml'
```

And (in the same logs):

```python-traceback
2024-06-15T09:03:37.8325369Z ERROR: Traceback (most recent call last):
2024-06-15T09:03:37.8326307Z   File "/GaNDLF/GANDLF/entrypoints/run.py", line 66, in _run
2024-06-15T09:03:37.8326861Z     main_run(
2024-06-15T09:03:37.8329112Z   File "/GaNDLF/GANDLF/cli/main_run.py", line 121, in main_run
2024-06-15T09:03:37.8329773Z     InferenceManager(
2024-06-15T09:03:37.8330508Z   File "/GaNDLF/GANDLF/inference_manager.py", line 69, in InferenceManager
2024-06-15T09:03:37.8331314Z     inference_loop(
2024-06-15T09:03:37.8332085Z   File "/GaNDLF/GANDLF/compute/inference_loop.py", line 89, in inference_loop
2024-06-15T09:03:37.8333459Z     assert file_to_load != None, "The 'best_file' was not found"
2024-06-15T09:03:37.8334690Z AssertionError: The 'best_file' was not found
```

VukW:
Hi folks! I was going just to comment how to fix the issue, but it seems it's easier to show the fix on the code itself.
https://github.com/benmalef/GaNDLF/pull/6

So, the root issue is you are using a relative `logging_config.yaml` path; it resolves to the current user working directory. So, if you run tests, the working directory is `./testing/`, while config is located at `./logging_config.yaml`.

To fix that we need to ensure that logging config path does not depend on user's workdir. The best solution is to add this file to the wheel and install it together with gandlf's module (`site-packages/GANDLF` in user's python env). At the same time, adding the file to the wheel is crucial as we want to distribute config together with pip package.

So, the solution includes:
(1) move `'logging_config.yaml` from repo root to `./GANDLF/` module source (only files located in module sources can be distributed with wheel)
(2) modify `setup.py`, `MANIFEST.in` to include this file to the package (required some `setup.py` cleaning beforehand, I've done that here https://github.com/mlcommons/GaNDLF/pull/890)
(3) modify logging configuration code to use file from package resources instead of usual relative path.

GANDLF/utils/gandlf_logger.py
```diff
@@ -0,0 +1,44 @@
...
+    Returns:
+        logging.Logger
+    """
+    with open(config_path, "r") as file:
+        config1 = yaml.safe_load(file)
+        logging.config.dictConfig(config1)
+
+    logging.captureWarnings(True)
+
+    return logging.getLogger(logger_name)
```
VukW, line 26: I just noticed you read configuration file & re-setups logging config each time you need to get a new logger.
What do you think if we split this onto two calls?
- `gandlf_logger_setup` would return None, just explicitly sets up the configuration. We call it explicitly just once per every CLI entrypoint. IMO, the best places are:
  - https://github.com/mlcommons/GaNDLF/blob/new-apis_v0.1.0-dev/GANDLF/entrypoints/cli_tool.py#L27 for `gandlf` cli tool
  - https://github.com/mlcommons/GaNDLF/blob/new-apis_v0.1.0-dev/GANDLF/entrypoints/run.py#L186 `opd_way`  functions for every `gandlf_*` script
  
- The second part is just to use `logging.getLogger(???)` anywhere in the code to get a logger. No need to reconfigure anything. Also, thirdparty libraries would also be covered.

VukW:
@benmalef You can just check this: https://github.com/benmalef/GaNDLF/pull/6 here I have implemented the fix already. My comment was actually an explanation of what exactly (and why) I did to fix the issue:)

benmalef:
> @benmalef You can just check this: [benmalef#6](https://github.com/benmalef/GaNDLF/pull/6) here I have implemented the fix already. My comment was actually an explanation of what exactly (and why) I did to fix the issue:)



> @benmalef You can just check this: [benmalef#6](https://github.com/benmalef/GaNDLF/pull/6) here I have implemented the fix already. My comment was actually an explanation of what exactly (and why) I did to fix the issue:)



> @benmalef You can just check this: [benmalef#6](https://github.com/benmalef/GaNDLF/pull/6) here I have implemented the fix already. My comment was actually an explanation of what exactly (and why) I did to fix the issue:)

aa ok. I will check it. Thanks a lot..!! :P

GANDLF/utils/gandlf_logger.py
benmalef: @VukW I tried to implement this in a [new PR](https://github.com/mlcommons/GaNDLF/pulls#issue-2387298821)

sarthakpati:
Fixed with #893

PR #889, opened 2024-06-28
Adding function to get augmentation transforms

szmazurek:
Fixes #888

## Proposed Changes
-  function for getting the augmentation transforms from params dict
- using that function in dataset constructor

## Checklist

- [X] [`CONTRIBUTING`](https://github.com/mlcommons/GaNDLF/blob/master/CONTRIBUTING.md) guide has been followed.
- [X] PR is based on the [current GaNDLF master ](https://docs.github.com/en/desktop/contributing-and-collaborating-using-github-desktop/keeping-your-local-repository-in-sync-with-github/syncing-your-branch-in-github-desktop?platform=windows).
- [X] Non-breaking change (does **not** break existing functionality): provide **as many** details as possible for _any_ breaking change.
- [X] Function/class source code documentation added/updated (ensure `typing` is used to provide type hints, including and not limited to using `Optional` if a variable has a pre-defined value).
- [X] Code has been [blacked](https://github.com/psf/black#usage) for style consistency and linting.
- [X] If applicable, version information [has been updated in GANDLF/version.py](https://github.com/mlcommons/GaNDLF/blob/master/GANDLF/version.py).
- [ ] If adding a git submodule, add to list of exceptions for black styling in [pyproject.toml](https://github.com/mlcommons/GaNDLF/blob/master/pyproject.toml) file.
- [X] [Usage documentation](https://github.com/mlcommons/GaNDLF/blob/master/docs) has been updated, if appropriate.
- [ ] Tests added or modified to [cover the changes](https://app.codecov.io/gh/mlcommons/GaNDLF); if coverage is reduced, please give explanation.
- [ ] If customized dependency installation is required (i.e., a separate `pip install` step is needed for PR to be functional), please ensure it is reflected in all the files that control the CI, namely: [python-test.yml](https://github.com/mlcommons/GaNDLF/blob/master/.github/workflows/python-test.yml), and all docker files [[1](https://github.com/mlcommons/GaNDLF/blob/master/Dockerfile-CPU),[2](https://github.com/mlcommons/GaNDLF/blob/devcontainer_build_fix/Dockerfile-CUDA11.6),[3](https://github.com/mlcommons/GaNDLF/blob/master/Dockerfile-ROCm)].

PR #890, opened 2024-06-29
Cleaning setup

VukW:
## Proposed Changes

- Right now `setup.py` places some GANDLF repo files to `site-packages/` of the user's python env. This cleaning helps to keep all gandlf files in `site-packages/GANDLF` so keeping user env tidy + removes other unused parts of `setup.py`
- Fix for #889

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,9 +32,7 @@ dockerfiles = [
 
 # Any extra files should be located at `GANDLF` module folder (not in repo root)
 extra_files = []
-toplevel_package_excludes = [
-    "testing*",
-]
+toplevel_package_excludes = ["testing*"]
 
 # specifying version for `black` separately because it is also used to [check for lint](https://github.com/mlcommons/GaNDLF/blob/master/.github/workflows/black.yml)
 black_version = "23.11.0"
--- a/setup.py
+++ b/setup.py
@@ -5,9 +5,6 @@
 
 import sys, re, os
 from setuptools import setup, find_packages
-from setuptools.command.install import install
-from setuptools.command.develop import develop
-from setuptools.command.egg_info import egg_info
 
 try:
     with open("README.md") as readme_file:
@@ -15,21 +12,6 @@ try:
 except Exception as error:
     readme = "No README information found."
     sys.stderr.write("Warning: Could not open '%s' due %s\n" % ("README.md", error))
-
-
-class CustomInstallCommand(install):
-    def run(self):
-        install.run(self)
-
-
-class CustomDevelopCommand(develop):
-    def run(self):
-        develop.run(self)
-
-
-class CustomEggInfoCommand(egg_info):
-    def run(self):
-        egg_info.run(self)
 
 
 try:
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -17,21 +14,6 @@
...
-class CustomDevelopCommand(develop):
-    def run(self):
-        develop.run(self)
-
-
-class CustomEggInfoCommand(egg_info):
-    def run(self):
-        egg_info.run(self)
-
-
```
VukW, line 34: These commands are just not used (I mean, they does not differ from default ones)

setup.py
```diff
@@ -47,25 +29,11 @@ def run(self):
     for item in os.listdir(os.path.dirname(os.path.abspath(__file__)))
     if (os.path.isfile(item) and item.startswith("Dockerfile-"))
 ]
-setup_files = ["setup.py", ".dockerignore", "pyproject.toml", "MANIFEST.in"]
-all_extra_files = dockerfiles + setup_files
-all_extra_files_pathcorrected = [os.path.join("../", item) for item in all_extra_files]
```
VukW, line 52: After installement, user receives this files directly in `site-packages/` folder. The reason is using `../` in the path; however, that is caused by the fact these files are located in repo root.

Good practice for building a wheel by `setup.py` (and, actually, the only possible one) is **not to add** these files to the wheel. Only files that are required for module should be added. Moreover, any file that should be added to user's `site-packages/GANDLF` module should be located in module source library also (`./GANDLF/` in our case, not in a repo root).

So, there are two options:
(1) move necessary files to `./GANDLF/` module source folder;
(2) just don't include them to the wheel at all.

I chose the second solution.

setup.py
```diff
@@ -47,25 +29,11 @@ def run(self):
...
-    "data",
-    "grad_clipping",
-    "losses",
-    "metrics",
-    "models",
-    "optimizers",
-    "schedulers",
-    "utils",
+    "testing*",
 ]
```
VukW, line 37: They are not a top-level packages anymore (as they are located **inside** top-level GANDLF package). However, the `testing` is a top-level (together with `GANDLF`); thus we need to exclude it from a wheel (right now a wheel installs a `testing` module also)

sarthakpati (approved):

sarthakpati (commented):
Tests failing.

> FAILED testing/test_full.py::test_generic_config_read - TypeError: affine() got an unexpected keyword argument 'degrees'

VukW:
Hi!
Concerning pytest failure - the failed test is actually brought by https://github.com/mlcommons/GaNDLF/pull/889/ - new augmentations. Let's proceed there, restored branch & PR

Concerning linter issues - I'll fix it in this PR

szmazurek:
Hey! 
I agree with your approach @VukW for the setup file. Regarding the augmentations - interesting thing, I did not receive any errors during tests for this pr. Will also take a look, if you find some cause of that earlier please let me know.

sarthakpati:
Thanks, folks!

PR #893, opened 2024-07-02
Add logging final version

### LINKED ISSUE #755: Prettify stdout

benmalef:
Fixes #755 

## Brief Description
This PR fixes the issues from [this PR](https://github.com/mlcommons/GaNDLF/pull/881#issue-2333481102)

I have tried to implement this [ref](https://github.com/mlcommons/GaNDLF/pull/881#discussion_r1661348706)

I  have **NOT** implemented the tqdm [ref](https://github.com/mlcommons/GaNDLF/pull/881#issuecomment-2152197420). I will create a separate PR. 

## Screenshots
This screenshot shows how logging messages are in the file. 

image

## Proposed Changes

- 

## Checklist

### DIFF
--- a/GANDLF/entrypoints/cli_tool.py
+++ b/GANDLF/entrypoints/cli_tool.py
@@ -3,7 +3,7 @@ import logging
 import click
 from .subcommands import cli_subcommands
 from GANDLF.entrypoints import append_copyright_to_help
-from GANDLF.utils import gandlf_logger_setup
+from GANDLF.utils import logger_setup
 from GANDLF import version
 
 
@@ -25,7 +25,7 @@ def gandlf(ctx, loglevel):
     ctx.ensure_object(dict)
     ctx.obj["LOGLEVEL"] = loglevel
     # setup_logging(loglevel)
-    gandlf_logger_setup()
+    logger_setup()
 
 
 # registers subcommands: `gandlf anonymizer`, `gandlf run`, etc.
--- a/GANDLF/entrypoints/cli_tool.py
+++ b/GANDLF/entrypoints/cli_tool.py
@@ -7,24 +7,13 @@ from GANDLF.utils import logger_setup
 from GANDLF import version
 
 
-def setup_logging(loglevel):
-    logging.basicConfig(level=loglevel.upper())
-
-
 @click.group()
 @click.version_option(version, "--version", "-v", message="GANDLF Version: %(version)s")
-@click.option(
-    "--loglevel",
-    default="INFO",
-    help="Set the logging level (DEBUG, INFO, WARNING, ERROR, CRITICAL)",
-)
 @click.pass_context  # Pass the context to subcommands
 @append_copyright_to_help
-def gandlf(ctx, loglevel):
+def gandlf(ctx):
     """GANDLF command-line tool."""
     ctx.ensure_object(dict)
-    ctx.obj["LOGLEVEL"] = loglevel
-    # setup_logging(loglevel)
     logger_setup()
 
 
--- a/GANDLF/entrypoints/cli_tool.py
+++ b/GANDLF/entrypoints/cli_tool.py
@@ -3,7 +3,7 @@ import logging
 import click
 from .subcommands import cli_subcommands
 from GANDLF.entrypoints import append_copyright_to_help
-
+from GANDLF.utils import gandlf_logger_setup
 from GANDLF import version
 
 
@@ -22,6 +22,7 @@ def setup_logging(loglevel):
 @append_copyright_to_help
 def gandlf(ctx, loglevel):
     """GANDLF command-line tool."""
+    logger = gandlf_logger_setup(__name__)
     ctx.ensure_object(dict)
     ctx.obj["LOGLEVEL"] = loglevel
     setup_logging(loglevel)
--- a/GANDLF/entrypoints/cli_tool.py
+++ b/GANDLF/entrypoints/cli_tool.py
@@ -22,10 +22,10 @@ def setup_logging(loglevel):
 @append_copyright_to_help
 def gandlf(ctx, loglevel):
     """GANDLF command-line tool."""
-    logger = gandlf_logger_setup(__name__)
     ctx.ensure_object(dict)
     ctx.obj["LOGLEVEL"] = loglevel
-    setup_logging(loglevel)
+    # setup_logging(loglevel)
+    gandlf_logger_setup()
 
 
 # registers subcommands: `gandlf anonymizer`, `gandlf run`, etc.
--- a/GANDLF/logging_config.yaml
+++ b/GANDLF/logging_config.yaml
@@ -7,7 +7,7 @@ filters:
     (): logging.Filter
     name: "py.warnings"
   info_only_filter:
-    (): gandlf_logger.InfoOnlyFilter
+    (): utils.gandlf_logger.InfoOnlyFilter
 handlers:
   stdoutHandler: # only display info level
     class: logging.StreamHandler
--- a/GANDLF/logging_config.yaml
+++ b/GANDLF/logging_config.yaml
@@ -11,7 +11,7 @@ formatters:
       CRITICAL: "bold_red"
   simple:
     (): colorlog.ColoredFormatter
-    format: "%(log_color)s%(asctime)s - %(name)s - %(levelname)s - %(message)s"
+    format: "%(log_color)s%(asctime)s - %(levelname)s - %(message)s"
     datefmt: "%Y-%m-%d %H:%M:%S"
 filters:
   warnings_filter:
--- a/GANDLF/logging_config.yaml
+++ b/GANDLF/logging_config.yaml
@@ -3,8 +3,15 @@ formatters:
   detailed:
     format: "%(asctime)s - %(name)s - %(levelname)s - %(module)s:%(funcName)s:%(lineno)d - %(message)s"
     datefmt: "%Y-%m-%d %H:%M:%S"
+    log_colors:
+      DEBUG: "white"
+      INFO: "green"
+      WARNING: "yellow"
+      ERROR: "red"
+      CRITICAL: "bold_red"
   simple:
-    format: "%(asctime)s - %(name)s - %(levelname)s - %(message)s"
+    (): colorlog.ColoredFormatter
+    format: "%(log_color)s%(asctime)s - %(name)s - %(levelname)s - %(message)s"
     datefmt: "%Y-%m-%d %H:%M:%S"
 filters:
   warnings_filter:
@@ -28,7 +35,6 @@ handlers:
     class: logging.handlers.RotatingFileHandler
     level: DEBUG
     formatter: detailed
-    filename: tmp/gandlf/gandlf.log
     maxBytes: 10485760
     backupCount: 2
 loggers: # you can add your customized logger
--- a/GANDLF/logging_config.yaml
+++ b/GANDLF/logging_config.yaml
@@ -2,6 +2,8 @@ version: 1
 formatters:
   detailed:
     format: "%(asctime)s - %(name)s - %(levelname)s - %(message)s"
+  detailed1:
+    format: "%(asctime)s - %(name)s - %(levelname)s - %(pathname)s:%(lineno)d - %(message)s"
 filters:
   warnings_filter:
     (): logging.Filter
@@ -12,20 +14,20 @@ handlers:
   stdoutHandler: # only display info level
     class: logging.StreamHandler
     level: INFO
-    formatter: detailed
+    formatter: detailed1
     filters: [info_only_filter]
     stream: ext://sys.stdout
   stderrHandler: # display warning and above messages
     class: logging.StreamHandler
     level: WARNING
-    formatter: detailed
+    formatter: detailed1
     stream: ext://sys.stderr
   rotatingFileHandler:
     class: logging.handlers.RotatingFileHandler
     level: DEBUG
-    formatter: detailed
+    formatter: detailed1
     filename: tmp/gandlf/gandlf.log
-    maxBytes: 51200
+    maxBytes: 10485760
     backupCount: 2
 loggers: # you can add your customized logger
   debug_logger:
--- a/GANDLF/logging_config.yaml
+++ b/GANDLF/logging_config.yaml
@@ -18,7 +18,7 @@ filters:
     (): logging.Filter
     name: "py.warnings"
   info_only_filter:
-    (): GANDLF.utils.gandlf_logger.InfoOnlyFilter
+    (): GANDLF.utils.gandlf_logging.InfoOnlyFilter
 handlers:
   stdoutHandler: # only display info level
     class: logging.StreamHandler
--- a/GANDLF/logging_config.yaml
+++ b/GANDLF/logging_config.yaml
@@ -7,7 +7,7 @@ filters:
     (): logging.Filter
     name: "py.warnings"
   info_only_filter:
-    (): GANDLF.gandlf_logger.InfoOnlyFilter
+    (): gandlf_logger.InfoOnlyFilter
 handlers:
   stdoutHandler: # only display info level
     class: logging.StreamHandler
--- a/GANDLF/logging_config.yaml
+++ b/GANDLF/logging_config.yaml
@@ -1,9 +1,11 @@
 version: 1
 formatters:
   detailed:
+    format: "%(asctime)s - %(name)s - %(levelname)s - %(module)s:%(funcName)s:%(lineno)d - %(message)s"
+    datefmt: "%Y-%m-%d %H:%M:%S"
+  simple:
     format: "%(asctime)s - %(name)s - %(levelname)s - %(message)s"
-  detailed1:
-    format: "%(asctime)s - %(name)s - %(levelname)s - %(pathname)s:%(lineno)d - %(message)s"
+    datefmt: "%Y-%m-%d %H:%M:%S"
 filters:
   warnings_filter:
     (): logging.Filter
@@ -14,18 +16,18 @@ handlers:
   stdoutHandler: # only display info level
     class: logging.StreamHandler
     level: INFO
-    formatter: detailed1
+    formatter: simple
     filters: [info_only_filter]
     stream: ext://sys.stdout
   stderrHandler: # display warning and above messages
     class: logging.StreamHandler
     level: WARNING
-    formatter: detailed1
+    formatter: detailed
     stream: ext://sys.stderr
   rotatingFileHandler:
     class: logging.handlers.RotatingFileHandler
     level: DEBUG
-    formatter: detailed1
+    formatter: detailed
     filename: tmp/gandlf/gandlf.log
     maxBytes: 10485760
     backupCount: 2
--- a/GANDLF/logging_config.yaml
+++ b/GANDLF/logging_config.yaml
@@ -1,0 +1,37 @@
+version: 1
+formatters:
+  detailed:
+    format: "%(asctime)s - %(name)s - %(levelname)s - %(message)s"
+filters:
+  warnings_filter:
+    (): logging.Filter
+    name: "py.warnings"
+  info_only_filter:
+    (): GANDLF.gandlf_logger.InfoOnlyFilter
+handlers:
+  stdoutHandler: # only display info level
+    class: logging.StreamHandler
+    level: INFO
+    formatter: detailed
+    filters: [info_only_filter]
+    stream: ext://sys.stdout
+  stderrHandler: # display warning and above messages
+    class: logging.StreamHandler
+    level: WARNING
+    formatter: detailed
+    stream: ext://sys.stderr
+  rotatingFileHandler:
+    class: logging.handlers.RotatingFileHandler
+    level: DEBUG
+    formatter: detailed
+    filename: tmp/gandlf/gandlf.log
+    maxBytes: 51200
+    backupCount: 2
+loggers: # you can add your customized logger
+  debug_logger:
+    level: DEBUG
+    handlers: [stdoutHandler, rotatingFileHandler, stderrHandler]
+    propagate: no
+root:
+  level: DEBUG
+  handlers: [stdoutHandler, rotatingFileHandler, stderrHandler]
--- a/GANDLF/logging_config.yaml
+++ b/GANDLF/logging_config.yaml
@@ -7,7 +7,7 @@ filters:
     (): logging.Filter
     name: "py.warnings"
   info_only_filter:
-    (): utils.gandlf_logger.InfoOnlyFilter
+    (): GANDLF.utils.gandlf_logger.InfoOnlyFilter
 handlers:
   stdoutHandler: # only display info level
     class: logging.StreamHandler
--- a/GANDLF/utils/gandlf_logger.py
+++ b/GANDLF/utils/gandlf_logger.py
@@ -3,6 +3,8 @@ import yaml
 from pathlib import Path
 from importlib import resources
 import colorlog
+import tempfile
+from GANDLF.utils import get_unique_timestamp
 
 
 def _flush_to_console():
@@ -23,13 +25,21 @@ def _flush_to_console():
     logging.root.addHandler(console_handler)
 
 
+def _create_tmp_log_file():
+    tmp_dir = Path(tempfile.gettempdir())
+    log_dir = Path.joinpath(tmp_dir, ".gandlf")
+    log_dir.mkdir(parents=True, exist_ok=True)
+    log_file = Path.joinpath(log_dir, get_unique_timestamp() + ".log")
+    _create_log_file(log_file)
+    return log_file
+
+
 def _create_log_file(log_file):
     log_file = Path(log_file)
     log_file.write_text("Starting GaNDLF logging session \n")
 
 
 def _save_logs_in_file(log_file, config_path):
-    _create_log_file(log_file)
     with resources.open_text("GANDLF", config_path) as file:
         config_dict = yaml.safe_load(file)
         config_dict["handlers"]["rotatingFileHandler"]["filename"] = str(log_file)
@@ -47,16 +57,18 @@ def gandlf_logger_setup(log_file=None, config_path="logging_config.yaml"):
     """
 
     logging.captureWarnings(True)
-
-    if log_file is None:  # flush logs
-        _flush_to_console()
-
-    else:  # create the log file
-        try:
+    try:
+        if log_file is None:  # create tmp file
+            log_tmp_file = _create_tmp_log_file()
+            _save_logs_in_file(log_tmp_file, config_path)
+            logging.info(f"The logs are saved in {log_tmp_file}")
+        else:  # create the log file
+            _create_log_file(log_file)
             _save_logs_in_file(log_file, config_path)
-        except Exception as e:
-            logging.error(f"log_file:{e}")
-            logging.warning("The logs will be flushed to console")
+    except Exception as e:
+        _flush_to_console()
+        logging.error(f"log_file:{e}")
+        logging.warning("The logs will be flushed to console")
 
 
 class InfoOnlyFilter(logging.Filter):
--- a/GANDLF/utils/gandlf_logger.py
+++ b/GANDLF/utils/gandlf_logger.py
@@ -2,7 +2,6 @@ import logging
 import yaml
 from pathlib import Path
 from importlib import resources
-import os
 import colorlog
 
 
--- a/GANDLF/utils/gandlf_logger.py
+++ b/GANDLF/utils/gandlf_logger.py
@@ -46,7 +46,7 @@ def _save_logs_in_file(log_file, config_path):
         logging.config.dictConfig(config_dict)
 
 
-def gandlf_logger_setup(log_file=None, config_path="logging_config.yaml"):
+def gandlf_logger_setup(log_file=None, config_path="logging_config.yaml") -> None:
     """
     It sets up the logger. Reads from logging_config.
 
--- a/GANDLF/utils/gandlf_logger.py
+++ b/GANDLF/utils/gandlf_logger.py
@@ -5,9 +5,7 @@ from pathlib import Path
 from importlib import resources
 
 
-def gandlf_logger_setup(
-     config_path="logging_config.yaml"
-) -> logging.Logger:
+def gandlf_logger_setup(config_path="logging_config.yaml") -> logging.Logger:
     """
     It sets up the logger. Read from logging_config.
     Args:
@@ -21,7 +19,6 @@ def gandlf_logger_setup(
     current_dir = Path.cwd()
     directory = Path.joinpath(current_dir, "tmp/gandlf")
     directory.mkdir(parents=True, exist_ok=True)
-
 
     with resources.open_text("GANDLF", config_path) as file:
         config_dict = yaml.safe_load(file)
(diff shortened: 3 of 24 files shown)


testing/test_full.py
```diff
@@ -3160,3 +3159,49 @@ def test_generic_data_split():
     sanitize_outputDir()
 
     print("passed")
+
+
+def test_gandlf_logging(capsys):
```
sarthakpati, line 3164: I suggest renaming this so that it gets invoked by the tests [ref].

```suggestion
def test_generic_logging(capsys):
```

sarthakpati (changes requested):
I would recommend adding a `Logging` section in the [documentation for extending GaNDLF](https://github.com/benmalef/GaNDLF/blob/add_logging_final_version/docs/extending.md) detailing this process and how a developer needs to use this correctly. For example, there should be 2 sub-sections to this:

1. What does someone need to do when they are _extending_ an existing function or class?
2. What does someone need to do when they are _adding_ a new function or class?
3. Anything else?

Am I missing something @VukW?

VukW:
@sarthakpati I agree it would be nice to mention logging in extension guide. However, the whole PR is created in the way that we hide all configuration from the developer - so developer almost doesn't need to bother about it. So, I believe there is not a big need to describe how to create loggers in new or modified code, but instead it would be useful to describe how logging is configured right now and how to log stuff sustainably
```
### Logging

#### Use loggers instead of print
We use native Python `logging` library for logs management. It is already configured, so if you are extending the code, please use loggers instead of `print` calls.
    ```
def my_new_cool_function(df: pd.DataFrame):
    logger = logging.getLogger(__name__)  # you can use any your own logger name or just pass a current file name
    logger.debug("Message for debug file only")
    logger.info("Hi GaNDLF user, I greet you in the CLI output")
    logger.error(f"A detailed message about any error if needed. Exception: {str(e)}, params: {params}, df shape: {df.shape}")
    # print("Hi GaNDLF user!")  # don't use prints please.
    ```

#### What and where is logged

GaNDLF logs are splitted into multiple parts:
- CLI output: only `info` messages are shown here
- debug file: ...
- errors file: ...
```

sarthakpati:
Sounds good, @VukW! Thank you for the explanation. 😄

GANDLF/logging_config.yaml
```diff
@@ -0,0 +1,39 @@
+version: 1
+formatters:
+  detailed:
+    format: "%(asctime)s - %(name)s - %(levelname)s - %(message)s"
+  detailed1:
+    format: "%(asctime)s - %(name)s - %(levelname)s - %(pathname)s:%(lineno)d - %(message)s"
```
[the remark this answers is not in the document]
benmalef: I have kept them because I don't know which format is more convenient. 
What do you believe is the best format? 
Also, it might be better to use different formats. For example, in stdout, we might not want a lot of information, but in the log file, it is useful to have extra information. 
Any thoughts? @VukW @sarthakpati

benmalef:
Hi guys @VukW, @sarthakpati, 
Thanks for the detailed review. 
I agree with it. I will do it.

GANDLF/logging_config.yaml
```diff
@@ -0,0 +1,39 @@
...
+    class: logging.StreamHandler
+    level: WARNING
+    formatter: detailed1
+    stream: ext://sys.stderr
+  rotatingFileHandler:
+    class: logging.handlers.RotatingFileHandler
+    level: DEBUG
+    formatter: detailed1
+    filename: tmp/gandlf/gandlf.log
+    maxBytes: 10485760
```
benmalef, line 30: @VukW What do you think is the best maxBytes for our log file?

benmalef:
@sarthakpati @VukW 
I added a  Logging section in the [documentation for extending GaNDLF](https://github.com/benmalef/GaNDLF/blob/add_logging_final_version/docs/extending.md#logging)

GANDLF/logging_config.yaml
VukW: I believe 1mb you've chosen is ok

testing/test_full.py
```diff
@@ -3160,3 +3159,49 @@ def test_generic_data_split():
     sanitize_outputDir()
 
     print("passed")
+
+
+def test_generic_logging(capsys):
```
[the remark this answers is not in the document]
VukW: No complains, looks good to me, man

GANDLF/logging_config.yaml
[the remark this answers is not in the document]
VukW: 1. IMO using different formatters (short and long) for CLI and debug logs is a good idea, let's do it
2. (detailed format) For me personally `pathname` looks too excessive - on your screenshot it takes almost all the space. I usually prefer to use `module` (python script file name, without the full path) or `module`+`funcName`. It seems enough for easy understanding where exactly message was written.

Example of how `module`+`funcName` instead of `pathname` look like:
`format: "%(asctime)s - %(name)s - %(levelname)s - %(module)s:%(funcName)s:%(lineno)d -  %(message)s"`
```
2024-07-04 11:49:51,148 - py.warnings - WARNING - warnings:_showwarnmsg:109 -  /mnt/c/Users/vykuk/repos/mlc/GaNDLF/check_logging_levels.py:15: UserWarning: a WARNING!!
  warnings.warn("a WARNING!!")  # debug log and stderr
2024-07-04 11:49:54,164 - my-logger - INFO - check_logging_levels:main:30 -  this is logging.INFO message
```
https://docs.python.org/3/library/logging.html#logrecord-attributes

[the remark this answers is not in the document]
benmalef: Yes, I agree. I like this format, is concise and smaller. I will change it. 
Thanks for your comments. :P

docs/extending.md
```diff
@@ -132,4 +132,40 @@ bash
 # continue from previous shell
 (venv_gandlf) $> coverage run -m pytest --device cuda; coverage report -m
 ```
+## Logging
+
+### Use loggers instead of print
+We use the native `logging` [library](https://docs.python.org/3/library/logging.html) for logs management.
+It is already configured. So, If you are extending the code, please use loggers instead of prints.
```
sarthakpati, line 139:
```suggestion
We use the native `logging` [library](https://docs.python.org/3/library/logging.html) for logs management. This gets automatically configured when GaNDLF gets launched. So, if you are extending the code, please use loggers instead of prints.
```

docs/extending.md
```diff
@@ -132,4 +132,40 @@ bash
...
+    logger.error(f"A detailed message about any error if needed. Exception: {str(e)}, params: {params}, df shape: {df.shape}")
+    # print("Hi GaNDLF user!")  # don't use prints please.
+```
+
+### What and where is logged
+
+GaNDLF logs are splitted into multiple parts:
+- CLI output: only `info` messages are shown here
+- debug file: all messages are shown 
+- stderr: display `warning,error,critical` messages
```
sarthakpati, line 157:
```suggestion
- stderr: display `warning`, `error`, or `critical` messages
```

GANDLF/utils/gandlf_logger.py
```diff
@@ -0,0 +1,44 @@
...
+    Args:
+        logger_name (str): logger name, the name should be the same in the logging_config
+        config_path (str): file path for the configuration
+    Returns:
+        logging.Logger
+    """
+
+    # create dir for storing the messages
+    current_dir = Path.cwd()
+    directory = Path.joinpath(current_dir, "tmp/gandlf")
```
sarthakpati, line 19: This can be prone to breaks. For example, if an HPC admin has done an installation of GaNDLF into an environment that requires elevated privileges to make changes (a very common form of structuring "frozen" virtual environments in HPCs), then logging would break. A much more robust solution would be to generate the logs in the `output_dir` of the `parameters` dict, which gets initialized during both training [[ref](https://github.com/mlcommons/GaNDLF/blob/master/GANDLF/compute/training_loop.py#L235)] and inference [[ref](https://github.com/mlcommons/GaNDLF/blob/master/GANDLF/inference_manager.py#L37)]. This (or a similar) option will almost invariably be present for any GaNDLF process, so we can use that. In case there isn't any such option, I would vote to flush to console rather than putting it into this directory. 

Thoughts @VukW?

docs/extending.md
```diff
@@ -132,4 +132,40 @@ bash
...
+
+All the logs are saved in the `tmp/gandlf` directory
```
sarthakpati, line 159: This might change based on how the previous comment gets addressed.

docs/extending.md
```diff
@@ -132,4 +132,40 @@ bash
...
+
+Example of log message
+```
+#format: "%(asctime)s - %(name)s - %(levelname)s - %(pathname)s:%(lineno)d - %(message)s"
+2024-07-03 13:05:51,642 - root - DEBUG - GaNDLF/GANDLF/entrypoints/anonymizer.py:28 - input_dir='.'
+```
+
+### Create your own logger
+You can create and configure your own logger in the `GANDLF\logging_config.yaml`.
```
sarthakpati, line 168:
```suggestion
You can create and configure your own logger by updating the file `GANDLF/logging_config.yaml`.
```

sarthakpati (changes requested):

GANDLF/utils/gandlf_logger.py
VukW: What we do have in params?
`anonymizer`: output folder
`collect_stats`: output folder
`config_generator`: output folder
`construct_csv`: output file
`debug_info`: none
`deploy`: output folder
`generate_metrics`: output file
`optimize_model`: None? (there is a input file `model.pth.tar`, where is it located? maybe we can save logs nearby?)
`patch_miner`: output folder
`preprocess`: output folder
`recover_config`: output file
`run`: output folder (`--model-dir` arg)
`split-csv`: output folder
`verify-install`: None

So we don't have a fixed path for all the commands altogether - it needs to be calculated dynamically. The issue is we fix a logging file location in the `logging_config`..

Maybe we can do a workaround? Instead of passing a different output folder params for different commands, make a common `--output-dir` param for the whole `gandlf` cli command, and pass it to the scripts if it requires such a dir? like,
```
gandlf --output-folder ./output run ...
gandlf --output-folder . recover-config ...
```
If so, then we can make say
- make a current path as output by default (`.`) so `recover-config`, `debug-info` and others to store logs right here
- for all other commands (like run, deploy etc) require a `--output-folder` param to be filled explicitly


Another possible solution is to store logs in temporary dir instead `/tmp/gandlf` for unix, `%TEMP%` for windows
https://docs.python.org/3/library/tempfile.html
https://stackoverflow.com/a/43418319

In both the solutions we have to define a log location dynamically while `gandlf_logger_setup` instead of hardcoding the path in `logging_config.yaml`

Personally I'm in favor of the second solution. How about you?

Third possible solution (a bit ugly as we setup logging __before__ normal commands args are parsed): unify all output dir args to the same name, say, `output-dir` so cli looks like this:
```
gandlf run --output-dir ./output/ ....
gandlf deploy --output-dir ./output/ ...
gandlf debug-info  # no output-dir here
```
Thus, while user enters `gandlf` cli, we extract this param from the context (if present) and use it as log folder destination. Instead, just use `.`

benmalef: Hi @VukW, how are you? 
I agree with you, the second solution is the best.
I will try to implement it. 
Thanks for your suggestions.!!!

sarthakpati: Thanks for pointing out the discrepancy for `optimize_model`, @VukW! I have added this something we need to fix in #895.

sarthakpati: > Thanks for pointing out the discrepancy for `optimize_model`, @VukW! I have added this something we need to fix in #895.

Update: this should get fixed with #896.

VukW: Hey @sarthakpati , that's great. However we still have `generate_metrics`, `construct_csv` and `recover_config` that do not have an output folder (just a path to output file). How do you think, can we unify this also to have an output folder? + we have `debug-info` + `verify-install` that do not have the output folder in their concept - so there's a question where to save gandlf logs for them

sarthakpati: I see 2 solutions:

1. How about this: let's default the location of the logs to `None`, in which case it flushes the logs to the console. In case the location is a directory, it writes the log there. This way, we don't have to change the invocation of `generate_metrics`, `construct_csv` and `recover_config` (which are relatively quick-running processes anyway).
2. Ensure a consistent CLI for all entrypoints that always has `output-path` as an input directory. This might create confusion, since CLI apps like `generate_metrics`, `construct_csv` and `recover_config` need to create a single file. For example, how do we handle the case where a user wants to pass different configurations to generate different metrics? 

From a software design perspective, I feel the "best practice" would be 2, but 1 works better from the perspective of usability. Thoughts, @VukW?

benmalef: I am not sure if it is a good idea. I am thinking about creating a new CLI command for logging. This way, if the user wants to store the logs, they can define the location. By default, we can flush the logs to the console.

VukW: @benmalef The issue is we want to have logs by default, not only when user explicitly turns it on. The reason is user might need to see a logs of a long three-days training job on the cluster that failed - so he (1) don't want to rerun the whole job to get the error, (2) might not be able to see the console output after job was killed.

So the whole goal of this PR is actually to add a persistent logging to the file so user can always reach to. `debug-info`, `generate-metrics`, `construct-csv` are actually exclusions from the general behavior: it's ok if we would not have log files for these commands as the commands are quick enough and usually user sees its console output anyway.

VukW: > How about this: let's default the location of the logs to None, in which case it flushes the logs to the console. In case the location is a directory, it writes the log there. This way, we don't have to change the invocation of generate_metrics, construct_csv and recover_config (which are relatively quick-running processes anyway).

🥇After thoughts, I believe, this your @sarthakpati  solution looks best to me. Maybe it's not so elegant from the point of code design, but it should be more comfortable and intuitive for users. For us it means we revert my previous proposal (where `gandlf_logger_setup` is called just once in `gandlf` cli command initialization) and instead run the configuration in the start of every "real" command code (right after cli is already called and cli params are already parsed)

🥈At the second option - same, but unify `--output-dir` param name, so we can extract it from the cli context & set up logging just once for all `gandlf` commands

sarthakpati: Agreed. We can have a follow-up PR where we specify `--log-file` and include this into the CLI apps. Sounds good, @VukW @benmalef?

benmalef: @sarthakpati Sounds good to me.

benmalef: > > How about this: let's default the location of the logs to None, in which case it flushes the logs to the console. In case the location is a directory, it writes the log there. This way, we don't have to change the invocation of generate_metrics, construct_csv and recover_config (which are relatively quick-running processes anyway).
> 
> 🥇After thoughts, I believe, this your @sarthakpati solution looks best to me. Maybe it's not so elegant from the point of code design, but it should be more comfortable and intuitive for users. For us it means we revert my previous proposal (where `gandlf_logger_setup` is called just once in `gandlf` cli command initialization) and instead run the configuration in the start of every "real" command code (right after cli is already called and cli params are already parsed)
> 
> 🥈At the second option - same, but unify `--output-dir` param name, so we can extract it from the cli context & set up logging just once for all `gandlf` commands


@VukW  If I get this right, these steps should be done? 
1. Change the ganldf_logger_setup()
```
def gandlf_logger_setup(log_dir=None)
   If (log_dir == None)
   # flush  the logs to the console 
   else:
  # store the logs in the log_dir
```
2. Call the `gandlf_logger_setup() ` in CLI "real" commands after parameters are parsed. 
For example, for the `collect_stats` command [ref](GANDLF/entrypoints/collect_stats.py): 

``` 
 args = parser.parse_args()
gandlf_logger_setup(args.outputdir)
 _collect_stats(args.modeldir, args.outputdir)
```

sarthakpati: I agree. However, let's do 2 in a separate PR so that it is cleaner.

docs/extending.md
sarthakpati: This still needs to change.

sarthakpati (changes requested):

GANDLF/utils/gandlf_logger.py
```diff
@@ -0,0 +1,61 @@
...
+    """
+    It sets up the logger. Reads from logging_config.
+    If log_dir is None, the logs are flashed to console.
+    Args:
+        log_dir (str): dir path for saving the logs
+        config_path (str): file path for the configuration
+
+    """
+
+    if log_dir == None:  # flash logs
```
sarthakpati, line 18: Per [pep-8 suggestions](https://peps.python.org/pep-0008/#programming-recommendations):
```suggestion
    if log_dir is None:  # flash logs
```

GANDLF/utils/gandlf_logger.py
```diff
@@ -0,0 +1,61 @@
...
+import colorlog
+
+
+def gandlf_logger_setup(log_dir=None, config_path="logging_config.yaml"):
+    """
+    It sets up the logger. Reads from logging_config.
+    If log_dir is None, the logs are flashed to console.
+    Args:
+        log_dir (str): dir path for saving the logs
+        config_path (str): file path for the configuration
```
sarthakpati, line 14: I would suggest to pass a full path to `log_file` instead of a `log_dir`, because if we create default file names, there is a good chance for us to overwrite/append different logs, which can make things tougher for the end user.

```suggestion
def gandlf_logger_setup(log_file=None, config_path="logging_config.yaml"):
    """
    It sets up the logger. Reads from logging_config.
    Args:
        log_file (str): dir path for saving the logs, defaults to `None`, at which time logs are flushed to console.
        config_path (str): file path for the configuration
```

GANDLF/utils/gandlf_logger.py
```diff
@@ -0,0 +1,61 @@
...
+
+    else:  # create the log file
+        output_dir = Path(log_dir)
+        Path(output_dir).mkdir(parents=True, exist_ok=True)
+        with resources.open_text("GANDLF", config_path) as file:
+            config_dict = yaml.safe_load(file)
+            config_dict["handlers"]["rotatingFileHandler"]["filename"] = str(
+                Path.joinpath(output_dir, "gandlf.log")
+            )
+            logging.config.dictConfig(config_dict)
```
sarthakpati, line 43: Here, I will suggest the following (pseudo-code):

```
try:
    write a single line to the `log_file` (something like `"Starting GaNDLF logging session"`).
except:
    # this means that the user does not have write access to the location given by `log_file`, so give that error, and tell the user that we are falling back to the default of flushing output to console
    call logging setup again but with `log_file` as `None`

**Review by sarthakpati — CHANGES_REQUESTED** [review]

File: GANDLF/utils/gandlf_logger.py (hunk 7 above)
Comment (VukW) [reply GANDLF/utils/gandlf_logger.py]: Hey folks!

> Agreed. We can have a follow-up PR where we specify --log-file and include this into the CLI apps. 

@sarthakpati , sorry, didn't get. Do you mean we add a new explicit param to specify log file path? It would be optional, right, and without default value? But in this case user who faces failure of his looongish training run won't have the log. I thought one of the main reasons for us to implement a better logging is to log debug details always (or, at least, by default) so user can debug his run in case of any issues?

Comment (sarthakpati) [reply GANDLF/utils/gandlf_logger.py]: > It would be optional, right

Yup!

> But in this case user who faces failure of his looongish training run won't have the log.

For now, until @benmalef puts the new CLI param in place. I think this is unfortunately going to be a case where a user needs to be cognizant of the fact that the logs are going to be long. Considering other factors (write access, default write locations, so on), I think we only have 2 options for the default:

1. Flush to console.
2. Write to a temporary file using the [tempfile module](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile).

I believe you would prefer 2, right, @VukW?

File: GANDLF/utils/gandlf_logger.py [hunk 13]
```diff
@@ -5,44 +5,58 @@
 import colorlog
 
 
-def gandlf_logger_setup(log_dir=None, config_path="logging_config.yaml"):
+def _flash_to_console():
```
sarthakpati, line 8: Typo
```suggestion
def _flush_to_console():
```

GANDLF/utils/gandlf_logger.py
```diff
@@ -5,44 +5,58 @@
...
-        output_dir = Path(log_dir)
-        Path(output_dir).mkdir(parents=True, exist_ok=True)
-        with resources.open_text("GANDLF", config_path) as file:
-            config_dict = yaml.safe_load(file)
-            config_dict["handlers"]["rotatingFileHandler"]["filename"] = str(
-                Path.joinpath(output_dir, "gandlf.log")
-            )
-            logging.config.dictConfig(config_dict)
+    if log_file is None:  # flash logs
+        _flash_to_console()
```
sarthakpati, line 52: Typo
```suggestion
    if log_file is None:  # flush logs
        _flush_to_console()
```

GANDLF/utils/gandlf_logger.py
sarthakpati: I am fine with flushing to console, but I understand what @VukW is saying about it being a pain. Do you think changing the default behavior to use a [temporary file](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile) would be quick, @benmalef? 

I am fine with 1 & 2 of your proposals, but 3 & 4 represent a significant departure from the current workflow of scientific computing software packages, and I feel needs discussion amongst the broader audience/developer group.

benmalef: @sarthakpati @VukW 
I tried to implement [here](https://github.com/benmalef/GaNDLF/pull/15)
Also, I tried to use the tempfile module, but I get some errors.

GANDLF/utils/gandlf_logger.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+def _save_logs_in_file(log_file, config_path):
+    with resources.open_text("GANDLF", config_path) as file:
+        config_dict = yaml.safe_load(file)
+        config_dict["handlers"]["rotatingFileHandler"]["filename"] = str(log_file)
+        logging.config.dictConfig(config_dict)
+
+
+def gandlf_logger_setup(log_file=None, config_path="logging_config.yaml"):
```
sarthakpati, line 49: Since this is an externally visible function, I would suggest adding the return to helping typing suggestions/hints.

```suggestion
def gandlf_logger_setup(log_file=None, config_path="logging_config.yaml") -> None:
```

GANDLF/utils/gandlf_logger.py
```diff
@@ -0,0 +1,87 @@
...
+            log_tmp_file = _create_tmp_log_file()
+            _save_logs_in_file(log_tmp_file, config_path)
+            logging.info(f"The logs are saved in {log_tmp_file}")
+        else:  # create the log file
+            _create_log_file(log_file)
+            _save_logs_in_file(log_file, config_path)
+    except Exception as e:
+        _flush_to_console()
+        logging.error(f"log_file:{e}")
+        logging.warning("The logs will be flushed to console")
```
sarthakpati, line 71: I would suggest the following execution order for clarity (also, `try` is no longer needed since the temp file should always have user-level write access):

```suggestion
    log_tmp_file = log_file
    if log_file is None:  # create tmp file
        log_tmp_file = _create_tmp_log_file()
        logging.info(f"The logs are saved in {log_tmp_file}")
    _create_log_file(log_tmp_file)
    _save_logs_in_file(log_tmp_file, config_path)
```

GANDLF/utils/gandlf_logger.py
```diff
@@ -0,0 +1,87 @@
...
+            "INFO": "green",
+            "WARNING": "yellow",
+            "ERROR": "red",
+            "CRITICAL": "bold_red",
+        },
+    )
+    console_handler = logging.StreamHandler()
+    console_handler.setFormatter(formatter)
+    logging.root.setLevel(logging.DEBUG)
+    logging.root.addHandler(console_handler)
```
sarthakpati, line 25: If the `try` block is removed from the module below, then this function is no longer needed, right?
```suggestion
```

sarthakpati (commented):
Minor comments, and we should be ready to merge! Thanks a ton for this, @benmalef!

GANDLF/utils/gandlf_logger.py
benmalef: Yes, you are right

GANDLF/utils/gandlf_logger.py
benmalef: yes if we don't want to flush to the console, the function is no longer needed.

GANDLF/utils/gandlf_logger.py
benmalef: Yes, this code is cleaner.  Thanks for the refactor.!

benmalef:
@sarthakpati I made the proposed changes...!! Thanks a lot for the suggestions.!!

GANDLF/utils/gandlf_logger.py
sarthakpati, line 1: Since it is already part of the `GANDLF` module, can you rename this file to just `logging.py`?

GANDLF/utils/gandlf_logger.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def _save_logs_in_file(log_file, config_path):
+    with resources.open_text("GANDLF", config_path) as file:
+        config_dict = yaml.safe_load(file)
+        config_dict["handlers"]["rotatingFileHandler"]["filename"] = str(log_file)
+        logging.config.dictConfig(config_dict)
+
+
+def gandlf_logger_setup(log_file=None, config_path="logging_config.yaml") -> None:
```
sarthakpati, line 29: Just a minor semantic change (since it is already part of the top-level `GANDLF` module). I am guessing this would have some other associated changes.

```suggestion
def logger_setup(log_file=None, config_path="logging_config.yaml") -> None:
```

sarthakpati (changes requested):
Minor semantic change. This PR should be good to merge after this.

GANDLF/utils/gandlf_logger.py
benmalef: I think naming the file 'logging' might be causing confusion because it has the same name as the library.

GANDLF/utils/gandlf_logger.py
benmalef: We can change it

GANDLF/utils/gandlf_logger.py
benmalef: A better name might be `gandlf_logging.py`

sarthakpati: I see. Okay, then, let's merge this if @VukW no longer has any feedback and we can debate the nomenclature separately.

GANDLF/utils/gandlf_logger.py
VukW: @benmalef @sarthakpati 
concerning (3) I believe it's both ok to make init once in `cli_tool` if possible (it was actually created to unite some general stuff like logging), as well as init logs in a separate commands if it's necessary.

concerning (4) and `gandlf init` - I agree with Sarthak that this would require a loot of changes (like, config management  / versioning / init), so I'd prefer not to implement this right now.

GANDLF/utils/gandlf_logger.py
VukW: Hey folks, I'd check the code detailly today

GANDLF/utils/gandlf_logger.py
sarthakpati: Thanks @VukW!

> concerning (3) I believe it's both ok to make init once in `cli_tool` if possible 

@benmalef I guess we can merge once this is done. Thanks!

benmalef: @sarthakpati  Perfect..!! Tomorrow I will start to implement the next PR (CLI commands)

GANDLF/entrypoints/cli_tool.py
```diff
@@ -24,7 +24,8 @@ def gandlf(ctx, loglevel):
     """GANDLF command-line tool."""
     ctx.ensure_object(dict)
     ctx.obj["LOGLEVEL"] = loglevel
-    setup_logging(loglevel)
+    # setup_logging(loglevel)
```
VukW, line 27: Let's remove redundant old loglevel stuff here

GANDLF/utils/gandlf_logging.py
```diff
@@ -0,0 +1,61 @@
+import logging
+import yaml
+from pathlib import Path
+from importlib import resources
+import tempfile
+from GANDLF.utils import get_unique_timestamp
+
+
+def _create_tmp_log_file():
+    tmp_dir = Path(tempfile.gettempdir())
```
VukW, line 10: Just to remind about weird tmp path in Mac OS: https://stackoverflow.com/questions/847850/cross-platform-way-of-getting-temp-directory-in-python/43418319#43418319
Up to you if we want to use this system one, or use a workaround to use `/tmp` in macos also. (I'm slightly in favor of using `/tmp`, but that's quite subjective and I never actually was a real mac os user)

GANDLF/utils/gandlf_logging.py
```diff
@@ -0,0 +1,61 @@
...
+    log_dir = Path.joinpath(tmp_dir, ".gandlf")
+    log_dir.mkdir(parents=True, exist_ok=True)
+    log_file = Path.joinpath(log_dir, get_unique_timestamp() + ".log")
```
VukW, line 13: 👍

GANDLF/utils/gandlf_logging.py
```diff
@@ -0,0 +1,61 @@
...
+    return log_file
+
+
+def _create_log_file(log_file):
+    log_file = Path(log_file)
+    log_file.write_text("Starting GaNDLF logging session \n")
+
+
+def _save_logs_in_file(log_file, config_path):
```
VukW, line 22: just a nitpick - this function is not saving any logs, instead just configures logging. Maybe rename it to smth more relevant? For ex,
```suggestion
def _configure_logging_with_logfile(log_file, config_path):
```

VukW (commented):
Checked the overall code, no any significant issues, looks good to me! Thanks, man @benmalef

GANDLF/utils/gandlf_logging.py
sarthakpati: > This is because on MacOS, i.e. Darwin, tempfile.gettempdir() and os.getenv('TMPDIR') return a value such as '/var/folders/nj/269977hs0_96bttwj2gs_jhhp48z54/T'; it is one that I do not always want.

Sorry, but I do not agree with this change. If someone is using macOS, they need to deal with their systems' idiosyncrasies, unless someone reports a proper issue.

sarthakpati:
The recent changes are good for me to merge. @VukW, if you are okay as well, let's merge this in and start the process of migrating the current master to an `old_api` branch and move on?

VukW (approved):
@sarthakpati Agree, looks good to me, let's merge it (just to remind, I believe @benmalef cannot merge it until your previous PR review result `Request Changes` is active)

PR #923, opened 2024-08-19
Assert messages to flush to the log file

### LINKED ISSUE #913: [FEATURE] How to ensure that `assert` messages are logged?
### Is your feature request related to a problem? Please describe.
Currently, all messages from the `assert` conditions flush to the console, which is fine but it does not present a unified logging interface.

### Describe the solution you'd like
It would be great if those messages could get flushed to the log file (perhaps in addition to the console).

### Describe alternatives you've considered
N.A.

### Additional context
Perhaps this could be a solution: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertLogs

benmalef:
Fixes #913 

## Proposed Changes

-  Updated gandlf_logging to flush the assert logs into the file using the  [sys.excepthook](https://docs.python.org/3/library/sys.html#sys.excepthook) 

## Checklist

### DIFF
--- a/GANDLF/utils/gandlf_logging.py
+++ b/GANDLF/utils/gandlf_logging.py
@@ -7,6 +7,7 @@ from GANDLF.utils import get_unique_timestamp
 import sys
 import traceback
 
+
 def _create_tmp_log_file():
     tmp_dir = Path(tempfile.gettempdir())
     log_dir = Path.joinpath(tmp_dir, ".gandlf")
@@ -26,6 +27,7 @@ def _configure_logging_with_logfile(log_file, config_path):
         config_dict["handlers"]["rotatingFileHandler"]["filename"] = str(log_file)
         logging.config.dictConfig(config_dict)
 
+
 def gandlf_excepthook(exctype, value, tb):
     if issubclass(exctype, AssertionError):
         logging.exception(
@@ -34,6 +36,7 @@ def gandlf_excepthook(exctype, value, tb):
         )
     else:
         sys.__excepthook__(exctype, value, tb)
+
 
 def logger_setup(log_file=None, config_path="logging_config.yaml") -> None:
     """
@@ -53,6 +56,7 @@ def logger_setup(log_file=None, config_path="logging_config.yaml") -> None:
     _create_log_file(log_tmp_file)
     _configure_logging_with_logfile(log_tmp_file, config_path)
     sys.excepthook = gandlf_excepthook
+
 
 class InfoOnlyFilter(logging.Filter):
     """
--- a/GANDLF/utils/gandlf_logging.py
+++ b/GANDLF/utils/gandlf_logging.py
@@ -28,7 +28,7 @@ def _configure_logging_with_logfile(log_file, config_path):
 
 
 def gandlf_excepthook(exctype, value, tb):
-    if issubclass(exctype, AssertionError):
+    if issubclass(exctype, Exception):
         logging.exception("Uncaught exception", exc_info=(exctype, value, tb))
     else:
         sys.__excepthook__(exctype, value, tb)
--- a/GANDLF/utils/gandlf_logging.py
+++ b/GANDLF/utils/gandlf_logging.py
@@ -4,7 +4,8 @@ from pathlib import Path
 from importlib import resources
 import tempfile
 from GANDLF.utils import get_unique_timestamp
-
+import sys
+import traceback
 
 def _create_tmp_log_file():
     tmp_dir = Path(tempfile.gettempdir())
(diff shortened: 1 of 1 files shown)


sarthakpati:
This is good to go, right, @benmalef?

sarthakpati:
@VukW do you have any feedback?

VukW:
Hey folks, from a first glance it seems to be a legit way to log an exception, but to tell truth I'm confused (I've never used `excepthook` previously tho). let me check it a bit more deeper, cos I'm not 100% sure it should be used in this way

sarthakpati:
> Hey folks, from a first glance it seems to be a legit way to log an exception, but to tell truth I'm confused (I've never used `excepthook` previously tho). let me check it a bit more deeper, cos I'm not 100% sure it should be used in this way

Any update, @VukW?

GANDLF/utils/gandlf_logging.py
```diff
@@ -26,6 +28,16 @@ def _configure_logging_with_logfile(log_file, config_path):
         logging.config.dictConfig(config_dict)
 
 
+def gandlf_excepthook(exctype, value, tb):
+    if issubclass(exctype, AssertionError):
```
VukW, line 32: Despite of the initial issue is only about Assertions, let's log absolutely any uncaught exceptions here?

GANDLF/utils/gandlf_logging.py
```diff
@@ -26,6 +28,16 @@ def _configure_logging_with_logfile(log_file, config_path):
...
+        logging.exception(
+            f"{exctype.__name__}: {value}\n{''.join(traceback.format_exception(exctype, value, tb))}",
+            exc_info=False,
+        )
```
VukW, line 36:
```suggestion
        logging.exception("Uncaught exception", exc_info=(exctype, value, tb))
```
`Logging` can pretty-print exceptions on its own. Ow do we need a special formatting?

VukW:
@benmalef does PR work for you? For me `logging.exception` is not catch to the debug file...

benmalef:
> @benmalef does PR work for you? For me `logging.exception` is not catch to the debug file...

@VukW How do you test it? 
 It was fine when I tested it on my PC, but I'll check it again.
 Is logging.exception generally not caught, or is this only an issue in the excepthook?
Then, maybe this isn't the right approach.

benmalef:
## Important Notes 

I have tried to use the [sys.excepthook](https://docs.python.org/3/library/sys.html#sys.excepthook) to change the behavior of the assert statement. When a AssertionError is raised, it should be flushed in the console and saved in the log file.  In  my PC , it works well, at least with simple situations, but @VukW has mentioned that it is not working properly.  So maybe is not the right approach. 

### Testing
I have tried to test it,  but the pytest has its implementation of managing the AssertionError. 
One solution maybe is  [this](https://stackoverflow.com/questions/29036354/custom-sys-excepthook-doesnt-work-with-pytest).  I tried it but I didn't figure it out.
Another way that I have tried is to change the assert statement behavior in pytest. I followed some of these steps [ref](https://stackoverflow.com/questions/58321991/is-it-possible-to-change-pytests-assert-statement-behaviour-in-python) , but I got the same result. 

To cancel the default behavior it can be done using [this](https://docs.pytest.org/en/stable/how-to/assert.html#disabling-assert-rewriting)

sarthakpati:
@benmalef thank you for looking into this. 

@VukW: it would be great if you could provide some feedback. Until then, I would recommend we keep this PR as draft.

sarthakpati:
> everything work exactly as expected now for me

I think with this, we can proceed with the merge. I'll wait for the tests to finish.

VukW:
@sarthakpati sure, i don't have any other comments except already mentioned ones

GANDLF/utils/gandlf_logging.py
sarthakpati: Should the check be the following, instead?

```python
if issubclass(exctype, Exception)
```

GANDLF/utils/gandlf_logging.py
```diff
@@ -4,6 +4,8 @@
 from importlib import resources
 import tempfile
 from GANDLF.utils import get_unique_timestamp
+import sys
+import traceback
```
sarthakpati, line 8: Fix codacy error.

```suggestion
```

GANDLF/utils/gandlf_logging.py
benmalef: Yes, If we want to log all the uncaught exceptions.

sarthakpati: Great, then let's make that change and get ready to merge! 😄

sarthakpati (approved):

PR #976, opened 2024-11-22
Pydantic config

### LINKED ISSUE #758: Improved configuration parsing
Currently, the configuration parsing is done by a single submodule [[ref](https://github.com/mlcommons/GaNDLF/blob/master/GANDLF/parseConfig.py)], which is creates issues with maintenance as more functionality is introduced.

Using a configuration validation mechanism that is more standardized would make the developer and user experience much better. Some examples are (and this is by no means an exhaustive list):

| **Package** |               **Documentation**               |              **Example**             |
|:-----------:|:---------------------------------------------:|:------------------------------------:|
|   Cerberus  |       https://docs.python-cerberus.org/       | https://stackoverflow.com/a/46626418 |
| PyDantic    | https://docs.pydantic.dev/latest/             | https://stackoverflow.com/a/61021183 |
| Marshmallow | https://marshmallow.readthedocs.io/en/stable/ | https://stackoverflow.com/a/63739747 |

N.A.

**Additional context**
It would make sense to eventually move GaNDLF's various functionalities to something like [Hydra](https://hydra.cc/docs/intro/), and this could potentially be a good starting point.

benmalef:
Fixes #758

## Proposed Changes

- This is the **first implementation** of parameters configuration using pydantic.

## Checklist

- [ ] The `logging` library is being used and no `print` statements are left.


GANDLF/utils/pydantic_config.py
sarthakpati, line 1: This file badly needs documentation and comments.

benmalef:
@sarthakpati, what do you think about this approach?

For me, it is better to return a [Parameters](https://github.com/mlcommons/GaNDLF/pull/976/files#diff-86879ad193ee72d868f22b06bfefd47e1a06db3e3cce77b59700b0272c3c0775) object with all the necessary parameters, and not a [dictionary](https://github.com/mlcommons/GaNDLF/pull/976/files#diff-cba3be123b0f7a243d4f96a16f196104c5afc3032d87d646154a9ec53d734192R738),
 but maybe it requires a lot of changes in the source code.

sarthakpati:
> better to return a Parameters object with all the necessary parameters, and not a dictionary

Can you provide a brief pro/con reasoning as to why this would be the case? Especially since changing to a different data structure would require significant amount of changes throughout the code.

benmalef:
> > better to return a Parameters object with all the necessary parameters, and not a dictionary
> 
> Can you provide a brief pro/con reasoning as to why this would be the case? Especially since changing to a different data structure would require significant amount of changes throughout the code.

I took a look at the source code, and it is required a lot of changes. So, it is not a good idea. I thought it might offer a better code experience, because Pydantic provides some extra features. 

So, as regards the workflow, is it something like this? 
image

szmazurek:
Hey there!
So, @sarthakpati asked for a few words from my side here. Generally, I think it is a super nice idea to use Pydantic here, as this can automate a lot of checks. I think the approach is ok, and I agree with your todo comments @benmalef in the Parameters class, that some of the fields should themselves be Pydantic models and define their own structure and incorporate ALL possible options that can be specified there (of course, if such list is finite). For example, for the loss function, we can define something like this:
`     
class LossFunction(BaseModel):
    name: Literal["ce","cel" ... ]  # (and so on for all the values available in the losses dict)
    reduction: Optional[Literal["mean", "sum"]] = "mean" # here we allow to specify only those we support in our losses impl and also provide default values
    some_other_param_our_losses_use : Optional[Literal["some_value_we_allow",...] = "some_default_value"
    
class Parameters:
      ....
      loss_function: LossFunction
`
Not sure if this will work if the loss is only a string without a name keyword, but you get the point. 
Having that in such form, the workflow can run as follows:
Open yaml and load it as a dict -> pass it to the Parameters object -> Parameters automatically validates if the values passed are correct and inserts defaults -> params = pydantic_params_object.model_dump() 
And this way we have the parameters in the dict for, validated and ready to use. Additionally, such models hierarchy naturally will become a kind of documentation, where you see all available params and values.
What you guys think? If I was unclear on something ping me, and we can discuss it further. Cheers!

sarthakpati:
Thanks for the detailed explanation, @szmazurek! I updated your answer with some edits to make it clearer for myself. 

@benmalef: thoughts?

benmalef:
> Thanks for the detailed explanation, @szmazurek! I updated your answer with some edits to make it clearer for myself. BTW, since you are working on the Lightning integration, this might be relevant since could perhaps be some overlap between that and this PR. Also, we should do 2 separate tags for them so that we have distinct points of references for the users.
> 
> @benmalef: any thoughts on what Szymon has put forth?

Thanks a lot for your explanation, I I agree with that. 

My only concern is that:
In this ([_parserConfig](https://github.com/mlcommons/GaNDLF/blob/53d38491abd4a02b7ff98373b1b549923f3282c2/GANDLF/config_manager.py#L113) file), it loads the YAML file and has some validations and some code to regulate or init the parameters. So, might be better to pass the load dict to Parameters after this. 

1.  Open yaml and load it as a dict. 
2.  Use the _parserConfig to manipulate the dict ( init or calculate some parameters.)
3. Pass the loaded dict it to the pydantic Parameters object
4. Parameters automatically validates if the values passed are correct and inserts defaults
5. params = pydantic_params_object.model_dump()

Any thoughts.. ? @szmazurek  @szmazurek

sarthakpati:
@benmalef: there seems to be 2 rounds of initialization in your proposed workflow (points 2 and 4). I think this is going to lead to confusion between both end users and developers, and we need to figure out a way to have all parameter initialization happening in a single place.

benmalef:
@szmazurek, @sarthakpati  Hello, I hope you are doing...!!
I have started again the parameter configuration using Pydantic. Also, I have created a [pull request](https://github.com/benmalef/GaNDLF/pull/33) in my repo. 
I would like your review/feedback on this. 

It is the first version. :P

szmazurek:
> @szmazurek, @sarthakpati Hello, I hope you are doing...!! I have started again the parameter configuration using Pydantic. Also, I have created a [pull request](https://github.com/benmalef/GaNDLF/pull/33) in my repo. I would like your review/feedback on this.
> 
> It is the first version. :P

Hey there!
I will take a look this week - if I do not do it till Friday please ping me here.

ATB

benmalef:
> This has been superceded by #986, right?

@sarthakpati yesss..!!

PR #986, opened 2025-02-12
Added configuration management using pydantic 

benmalef:
add Pydantic configuration

Fixes #ISSUE_NUMBER

## Proposed Changes

- Add Pydantic configuration
- Refactor the config_manager
## Checklist

- [ ] The `logging` library is being used and no `print` statements are left.


szmazurek:
> @szmazurek could you please take a first pass?

yup, will do in like 1hr or tomorrow morning

GANDLF/Configuration/Parameters/default_parameters.py
```diff
@@ -0,0 +1,66 @@
...
+    medcam_enabled: bool = Field(
+        default=False, description="Enable interpretability via medcam."
+    )
+    save_training: bool = Field(
+        default=False, description="Save outputs during training."
+    )
+    save_output: bool = Field(
+        default=False, description="Save outputs during validation/testing."
+    )
+    in_memory: bool = Field(default=False, description="Pin data to CPU memory.")
```
szmazurek, line 21: What does it mean to "pin data to cpu/gpu memory"? Also, is the 'in_memory' really enforcing a page-lock on memory storing given chunk of data or just keeps it all in RAM?

GANDLF/Configuration/Parameters/default_parameters.py
```diff
@@ -0,0 +1,66 @@
...
+        default=False,
+        description="Enable memory-saving mode. If enabled, resize/resample will save files to disk.",
+    )
+    print_rgb_label_warning: bool = Field(
+        default=True, description="Print a warning for RGB labels."
+    )
+    data_postprocessing: Union[dict, set] = Field(
+        default={}, description="Default data postprocessing configuration."
+    )
+    grid_aggregator_overlap: str = Field(
```
szmazurek, line 57: What kind of other options can we have here? I believe this cannot be an arbitraty string, therefore it should be an optional literal here with available strings?

GANDLF/Configuration/Parameters/model_parameters.py
```diff
@@ -0,0 +1,68 @@
...
+
+# You can define new parameters for model here. Please read the pydantic documentation.
+# It allows extra fields in model dict.
+class Model(BaseModel):
+    model_config = ConfigDict(
+        extra="allow"
+    )  #  it allows extra fields in the model dict
+    dimension: Optional[int] = Field(description="Dimension.")
+    architecture: Union[ARCHITECTURE_OPTIONS, dict] = Field(description="Architecture.")
+    final_layer: str = Field(description="Final layer.")
```
szmazurek, line 23: Here we are also limited to certain amount of acceptable values - leveraging literal seems like good option.

GANDLF/Configuration/Parameters/model_parameters.py
```diff
@@ -0,0 +1,68 @@
...
+    )  # default is 32
+    class_list: Union[list, str] = Field(default=[], description="Class list.")
+    num_channels: Optional[int] = Field(
+        description="Number of channels.",
+        validation_alias=AliasChoices(
+            "num_channels", "n_channels", "channels", "model_channels"
+        ),
+        default=3,
+    )  # TODO: check it
+    type: Optional[str] = Field(description="Type of model.", default="torch")
```
szmazurek, line 38: Should it also be literal? Probably options are torch, openvino? @sarthakpati am I right?

GANDLF/Configuration/Parameters/model_parameters.py
```diff
@@ -0,0 +1,68 @@
...
+    data_type: str = Field(description="Data type.", default="FP32")
```
szmazurek, line 39: Is this true that we support such field in the config and it really influences anything in base gandlf? I tought that precision is chaning only when amp is enabled

GANDLF/Configuration/Parameters/model_parameters.py
```diff
@@ -0,0 +1,68 @@
...
+    save_at_every_epoch: bool = Field(default=False, description="Save at every epoch.")
+    amp: bool = Field(default=False, description="Amplifier.")
```
szmazurek, line 41: amp stands for automatic mixed precision, not amplifier

GANDLF/Configuration/Parameters/nested_training_parameters.py
```diff
@@ -0,0 +1,24 @@
...
+    )
+    testing: int = Field(
+        default=-5,
+        description="this controls the number of testing data folds for final model evaluation; [NOT recommended] to disable this, use '1'",
+    )
+    validation: int = Field(
+        default=-5,
+        description="this controls the number of validation data folds to be used for model *selection* during training (not used for back-propagation)",
+    )
+    proportional: Optional[bool] = Field(default=None)
```
szmazurek, line 18: what does this parameter do? also, if it's boolean, can't we set default as False?

GANDLF/Configuration/Parameters/nested_training_parameters.py
```diff
@@ -0,0 +1,24 @@
...
+from typing_extensions import Self, Optional
+
+
+class NestedTraining(BaseModel):
+    stratified: bool = Field(
+        default=False,
+        description="this will perform stratified k-fold cross-validation but only with offline data splitting",
+    )
+    testing: int = Field(
+        default=-5,
```
szmazurek, line 11: Open question - are there any limits to values that can be set in this field? Like, what happens if I set testing to 10? If there are limits, maybe it's worth to include possible ranges when defining this field, what do you think ? @sarthakpati @benmalef

GANDLF/Configuration/Parameters/patch_sampler.py
```diff
@@ -0,0 +1,8 @@
+from pydantic import BaseModel, Field
+
+
+class PatchSampler(BaseModel):
+    type: str = Field(default="uniform")
```
szmazurek, line 5: are there any other options available for type and padding_mode? if so, maybe we should use literal here?

GANDLF/Configuration/Parameters/scheduler_parameters.py
```diff
@@ -0,0 +1,17 @@
...
+
+# It allows extra parameters
+class Scheduler(BaseModel):
+    model_config = ConfigDict(extra="allow")
+    type: TYPE_OPTIONS = Field(
+        description="triangle/triangle_modified use LambdaLR but triangular/triangular2/exp_range uses CyclicLR"
+    )
+    # min_lr: 0.00001, #TODO: this should be defined ??
+    # max_lr: 1, #TODO: this should be defined ??
+    step_size: float = Field(description="step_size", default=None)
```
szmazurek, line 17: I think we need to think on different classes that would allow for definition of params for separate schedulers - for example, if we use a scheduler which does reduce on plateau, then we need a field to define tracked metric. Not really sure how to implement that nicely tho. Other approach is define all possible fields that any scheduler can take and later provide validation logic which takes care of conditionality - i.e if `reduce_on_plateau` type chosen, then we require `monitor` field

GANDLF/Configuration/Parameters/user_defined_parameters.py
```diff
@@ -0,0 +1,125 @@
...
+
+class InferenceMechanism(BaseModel):
+    grid_aggregator_overlap: Literal["crop", "average"] = Field(default="crop")
+    patch_overlap: int = Field(default=0)
+
+
+class UserDefinedParameters(DefaultParameters):
+    version: Version = Field(
+        default=Version(minimum=version("GANDLF"), maximum=version("GANDLF")),
+        description="Whether weighted loss is to be used or not.",
```
szmazurek, line 45: Descritpion is not valid I believe :P

GANDLF/Configuration/Parameters/user_defined_parameters.py
```diff
@@ -0,0 +1,125 @@
...
+    )
+    patch_size: Union[list[Union[int, float]], int, float] = Field(
+        description="Patch size."
+    )
+    model: Model = Field(..., description="The model to use. ")
```
szmazurek, line 50: should be list of avaiable strings no?

GANDLF/Configuration/Parameters/user_defined_parameters.py
```diff
@@ -0,0 +1,125 @@
...
+    ]
+    nested_training: NestedTraining = Field(description="Nested training.")
+    parallel_compute_command: str = Field(
+        default="", description="Parallel compute command."
+    )
+    scheduler: Union[str, Scheduler] = Field(
+        description="Scheduler.", default=Scheduler(type="triangle_modified")
+    )
+    optimizer: Union[str, Optimizer] = Field(
+        description="Optimizer.", default=Optimizer(type="adam")
```
szmazurek, line 70: Question about naming - here I first assumed we are initializing real torch optimizer - perhaps the names of config classes should be suffixed with `Config`/`params`?

GANDLF/Configuration/Parameters/user_defined_parameters.py
```diff
@@ -0,0 +1,125 @@
...
+    )  # TODO: Check it again for (opt)
+    patch_sampler: Union[str, PatchSampler] = Field(
+        description="Patch sampler.", default=PatchSampler()
+    )
+    inference_mechanism: InferenceMechanism = Field(
+        description="Inference mechanism.", default=InferenceMechanism()
+    )
+    data_postprocessing_after_reverse_one_hot_encoding: dict = Field(
+        description="data_postprocessing_after_reverse_one_hot_encoding.", default={}
```
szmazurek, line 79: Any options we support here? Not sure if this can be an arbitrary dict defined by user

GANDLF/Configuration/Parameters/user_defined_parameters.py
```diff
@@ -0,0 +1,125 @@
...
+    )
+    differential_privacy: Any = Field(description="Differential privacy.", default=None)
```
szmazurek, line 81: Can it be just a boolean field?

GANDLF/Configuration/Parameters/user_defined_parameters.py
```diff
@@ -0,0 +1,125 @@
...
+    # TODO: It should be defined with a better way (using a BaseModel class)
+    data_preprocessing: Annotated[
+        dict,
+        Field(description="Data preprocessing."),
+        AfterValidator(validate_data_preprocessing),
+    ] = {}
+    # TODO: It should be defined with a better way (using a BaseModel class)
```
szmazurek, line 88: I agree with the comments, it would allow for a lot of clarity

GANDLF/Configuration/utils.py
```diff
@@ -0,0 +1,67 @@
...
+        markdown.append(
+            f"| `{field_name}` | `{field_type}` | {description} | `{default}` |"
+        )
+
+    # Write to file
+    with open(file_path, "w", encoding="utf-8") as file:
+        file.write("\n".join(markdown))
+
+
+def initialize_key(
```
szmazurek, line 42: do we need such utility? Meaning, if there are default parameters to be set, ideally they are defined via pydantic and automatically populated if user did not set them explicitly

setup.py
```diff
@@ -85,6 +87,7 @@
     "openslide-bin",
     "openslide-python==1.4.1",
     "lion-pytorch==0.2.2",
+    "pydantic",
```
szmazurek, line 90: I would fix version, maybe in the future releases there are going to be some breaking changes of pydantic (little chance but I am paranoid a little)

GANDLF/Configuration/Parameters/default_parameters.py
benmalef: I took these default parameters https://github.com/mlcommons/GaNDLF/blob/3bfe133278f2b0b2b5322934c025f667205e2ebe/GANDLF/config_manager.py#L16
and tried to manipulate them using Pydantic.

GANDLF/Configuration/Parameters/model_parameters.py
```diff
@@ -0,0 +1,68 @@
...
+NORM_TYPE_OPTIONS = Literal["batch", "instance", "none"]
+
+
+# You can define new parameters for model here. Please read the pydantic documentation.
+# It allows extra fields in model dict.
+class Model(BaseModel):
+    model_config = ConfigDict(
+        extra="allow"
+    )  #  it allows extra fields in the model dict
+    dimension: Optional[int] = Field(description="Dimension.")
```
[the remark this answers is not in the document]
benmalef: If the user doesn't define the dimension, it is calculated in the `validate_patch_size(patch_size, dimension)` function in the validators file using the patch size.

Hmm, yes, maybe it should only accept 2 or 3. I will change it and update the description.

GANDLF/Configuration/Parameters/model_parameters.py
benmalef: Yesss, you are right.  I change it . :P

GANDLF/Configuration/Parameters/model_parameters.py
benmalef: I found it in the config_manager file and tried to manipulate it using pydantic . https://github.com/mlcommons/GaNDLF/blob/3bfe133278f2b0b2b5322934c025f667205e2ebe/GANDLF/config_manager.py#L594

GANDLF/Configuration/Parameters/nested_training_parameters.py
benmalef: I don't what does it, but i found it here
https://github.com/mlcommons/GaNDLF/blob/3bfe133278f2b0b2b5322934c025f667205e2ebe/GANDLF/config_manager.py#L638
So, I manipulate it using the def validate_nested_training(self) with the decorator model_validator.  
I donlt know what is the default value. We may set the default as False.

GANDLF/Configuration/Parameters/nested_training_parameters.py
benmalef: Hmm I don't know. Again, I found this value in the config_manager 
https://github.com/mlcommons/GaNDLF/blob/3bfe133278f2b0b2b5322934c025f667205e2ebe/GANDLF/config_manager.py#L642

GANDLF/Configuration/Parameters/user_defined_parameters.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+class Version(BaseModel):  # TODO: Maybe should be to another folder
+    minimum: str
+    maximum: str
+
+    @model_validator(mode="after")
+    def validate_version(self) -> Self:
+        if version_check(self.model_dump(), version_to_check=version("GANDLF")):
+            return self
```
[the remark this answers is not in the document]
benmalef: Yes, we will rise an assertion error in the version_check fun. 
https://github.com/mlcommons/GaNDLF/blob/3bfe133278f2b0b2b5322934c025f667205e2ebe/GANDLF/utils/generic.py#L91

GANDLF/Configuration/utils.py
benmalef: Yes, you are right.

The best way is not to use it, but some parameters have complex logic, and I need to figure out how to implement them using Pydantic.

So, I use it for some parameters as they are in the config_manager.

For example, for the data_augmentation parameter, I use it in the `validate_data_augmentation(value, patch_size)` function in the validators file.

GANDLF/Configuration/Parameters/scheduler_parameters.py
benmalef: hmm, if we could define different classes for each scheduler it would be great.  We should think about it.

setup.py
benmalef: Done. Updated the Pydantic in the  2.10.6 version.

GANDLF/Configuration/Parameters/user_defined_parameters.py
benmalef: I took this list https://github.com/mlcommons/GaNDLF/blob/3bfe133278f2b0b2b5322934c025f667205e2ebe/GANDLF/models/__init__.py#L41

and I define it as literal in the model architecture parameter.

GANDLF/Configuration/Parameters/model_parameters.py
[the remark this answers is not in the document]
benmalef: Here is the function https://github.com/benmalef/GaNDLF/blob/8837b7ccd25b747c7cbe4faaa77226a631febd85/GANDLF/Configuration/Parameters/validators.py#L132. in my branch. 

I tried to manipulate this code using the Pydantic.  https://github.com/mlcommons/GaNDLF/blob/3bfe133278f2b0b2b5322934c025f667205e2ebe/GANDLF/config_manager.py#L136

GANDLF/Configuration/Parameters/user_defined_parameters.py
benmalef: hahah.... I changed :P

GANDLF/Configuration/Parameters/user_defined_parameters.py
benmalef: what about `Optimizer_config` ?

GANDLF/Configuration/Parameters/user_defined_parameters.py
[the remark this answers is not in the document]
szmazurek: Ah, I see, thanks!

GANDLF/Configuration/Parameters/user_defined_parameters.py
szmazurek: Right, missed it

GANDLF/Configuration/Parameters/default_parameters.py
szmazurek: Regarding the comment about OOM with pin_memory (not really related to the pr tho, just in general) - pin memory may go OOM when the program runs at nearly full RAM utilization available within the machine, as page-locking will prevent parts of memory from being swapped, but I would advocate for allowing the user to try that option too (I did it in the lightning port)

benmalef:
Hey Guys @sarthakpati @szmazurek, 
I have a question about testing. I want to test the Pydantic configuration, and the approach I’m considering is to test each model class separately. Does it make sense to create different GitHub workflows for it , or should I write the tests directly in test_full.py

sarthakpati:
> test each model class separately

Is there a way to combine these tests?

> Does it make sense to create different GitHub workflows for it.

Sure. I would recommend naming these tests as `tests_${testing_type}_generic_config` so that it gets correctly called by the workflow [[ref](https://github.com/mlcommons/GaNDLF/blob/master/.github/workflows/python-test.yml#L26)].

GANDLF/Configuration/Parameters/user_defined_parameters.py
benmalef: done

GANDLF/Configuration/Parameters/scheduler_parameters.py
benmalef: @sarthakpati @szmazurek Can you take a look .? 

I tried to create separate class for each scheduler.   [Here](https://github.com/mlcommons/GaNDLF/pull/986/files#diff-a5820da3a5e1e2407d46252bee336d133854609ae7114d0c725fd341ca01dc34:~:text=def%20combine_models(base_model%3A%20Type%5BBaseModel%5D%2C%20extra_model%3A%20Type%5BBaseModel%5D)%3A) is the function that combines the scheduler models. The SchedulerConfig is the "base class" and has the type field. The parameters from other scheduler classes will be defined based on the type.  So, I have created this [dict](https://github.com/mlcommons/GaNDLF/pull/986/files#diff-a40aba76209b992066119bfc2844034201f3ff726b9d94635f7b93b67323bdf2:~:text=schedulers_dict_config) that contains all the scheduler config classes.  

Thanks.

GANDLF/Configuration/Parameters/user_defined_parameters.py
benmalef: done

GANDLF/Configuration/Parameters/nested_training_parameters.py
benmalef: done

GANDLF/Configuration/Parameters/model_parameters.py
benmalef: done

GANDLF/Configuration/Parameters/model_parameters.py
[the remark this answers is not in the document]
benmalef: done

GANDLF/Configuration/Parameters/nested_training_parameters.py
benmalef: done

GANDLF/configuration/scheduler_config.py
```diff
@@ -0,0 +1,90 @@
...
+    gamma: float = Field(default=0.1)
+    mode: Literal["min", "max"] = Field(default="min")
+    factor: float = Field(default=0.1)
+    patience: int = Field(default=10)
+    threshold: float = Field(default=0.0001)
+    cooldown: int = Field(default=0)
+    threshold_mode: Literal["rel", "abs"] = Field(default="rel")
+
+
+class warmupcosineschedule_config(BaseModel):
```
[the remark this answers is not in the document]
benmalef: Yes, you are right. :P
 I renamed.

GANDLF/configuration/user_defined_config.py
```diff
@@ -0,0 +1,130 @@
...
+        description="Inference mechanism.", default=InferenceMechanismConfig()
+    )
+    data_postprocessing_after_reverse_one_hot_encoding: dict = Field(
+        description="data_postprocessing_after_reverse_one_hot_encoding.", default={}
+    )
+    differential_privacy: Union[bool, DifferentialPrivacyConfig] = Field(
+        description="Differential privacy.", default=None
+    )
+    # TODO: It should be defined with a better way (using a BaseModel class)
+    data_preprocessing: Annotated[
```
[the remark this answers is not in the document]
benmalef: done

benmalef:
@szmazurek  @sarthakpati I think the PR is ready.😁 Can you please take a quick look.?

szmazurek (dismissed):
Thanks for the changes, looks solid.

sarthakpati (approved):

